PR #42, opened 2021-05-13
Fix search filter value type

gregnr:
# Background
Currently the search endpoint only supports filtering by the `'object'` property with possible values of `'page'` or `'database'`.

To do this with the SDK, you would do:
```typescript
const response = await notion.search({
  filter: {
    property: 'object',
    value: 'database',
  },
});
```

# Problem
The current type for `value` is the string literal `'object'` instead of `'page' | 'database'` which prevents TypeScript from compiling.

# Solution
Fixes the `SearchFilter` `value` type to be `'page' | 'database'`.

### DIFF
--- a/src/api-types.ts
+++ b/src/api-types.ts
@@ -316,7 +316,7 @@ export interface SearchSort {
 }
 
 export interface SearchFilter {
-  value: 'object',
+  value: 'page' | 'database',
   property: 'object';
 }
 


aoberoi:
@gregnr amazing, thanks for finding this. We won't be able to accept your PR until we have our Contributor License Agreement prepared. We certainly appreciate the help and look forward to accepting PRs soon.

I'll leave this open for now, and come back with an update when we're ready.

kcc989:
@aoberoi is there any chance you could open up your own PR and fix this? This is such a simple fix, but it is making it difficult to use the search endpoint the way I'd like right now.

gregnr:
@kcc989 My current workaround is:

```typescript
const response = await notion.search({
  filter: {
    property: 'object',
    value: 'database' as any,
  },
});
```

kcc989:
Ah that's a good call. I'm still new to typescript so that didn't occur to me. Thanks for your help!

blackmad:
thanks for pointing this out! fixed in https://github.com/makenotion/notion-sdk-js/commit/fc6defcef89d233f901c54f91ef8821fd73c99a7 so closing for now.

PR #57, opened 2021-05-14
Add default args for 'list' operations

btorresgil:
List users and list databases don't require and arguments and the
examples in the documentation don't provide any arguments, for example:

    notion.users.list()

Source: https://github.com/makenotion/notion-sdk-js#usage

However, when using this function in TypeScript, this example produces an
error:

> Expected 1 arguments, but got 0.ts(2554)
> Client.d.ts(84, 16): An argument for 'args' was not provided.

This is because `list()` requires an argument, even though all the properties of that argument are optional. To workaround the error, the developer has to provide an empty object:

    notion.users.list({})

This change sets a default argument of an empty object (`{}`) so the examples in the documentation work when using TypeScript and an argument is no longer needed.

Thanks for your great work on this API and SDK!

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -149,7 +149,7 @@ export default class Client {
     /**
      * List databases
      */
-    list: (args: WithAuth<DatabasesListParameters>): Promise<DatabasesListResponse> => {
+    list: (args: WithAuth<DatabasesListParameters> = {}): Promise<DatabasesListResponse> => {
       return this.request<DatabasesListResponse>({
         path: databasesList.path(),
         method: databasesList.method,
@@ -244,7 +244,7 @@ export default class Client {
     /**
      * List all users
      */
-    list: (args: WithAuth<UsersListParameters>): Promise<UsersListResponse> => {
+    list: (args: WithAuth<UsersListParameters> = {}): Promise<UsersListResponse> => {
       return this.request<UsersListResponse>({
         path: usersList.path(),
         method: usersList.method,


aoberoi:
@btorresgil Thanks for the fix!

We won't be able to accept your PR until we have our Contributor License Agreement prepared. We certainly appreciate the help and look forward to accepting PRs soon.

I'll leave this open for now, and come back with an update when we're ready.

Note: the API call still works as advertised at runtime, but @btorresgil is totally right that the type check fails.

blackmad:
[CLA assistant check](https://cla-assistant.io/makenotion/notion-sdk-js?pullRequest=57) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/makenotion/notion-sdk-js?pullRequest=57) before we can accept your contribution.<br/>

<hr/>

<sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/makenotion/notion-sdk-js?pullRequest=57) it.</sub>

blackmad:
thank you so much for this PR - the notion team fixed this grouped with a bunch of other fixes and will be releasing it later today, sorry we couldn't include your patch!

btorresgil:
@blackmad, thanks for the update. No problem, I know things are moving fast over there. Just glad to be part of the fun! Best of luck, guys!

PR #80, opened 2021-05-20
clean output directory "build" before building

mukeshkmr776:
Everytime a build is created, the output directory "build" should be cleaned.
This prevents from uncesessary/deleted files from last build and also helpful during development.


mukeshkmr776:
@blackmad  Please have a look on this PR.
First PR :)

PR #81, opened 2021-05-20
Apply prettier

justjake:
Add prettier and apply to the codebase. This should keep the codebase looking nice.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,13 +2,11 @@
 
 module.exports = {
   root: true,
-  parser: '@typescript-eslint/parser',
-  plugins: [
-    '@typescript-eslint'
-  ],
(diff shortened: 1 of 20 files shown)


blackmad:
why "don't use tabs" when our internal repo does?

blackmad:
other thing to clean up is that "//comments are formatted without a space after the double slash"

justjake:
@blackmad i think tabs inside markdown doc examples is pretty iffy, it indents things a lot and makes them hard to read. In a code editor it’s easy to configure tabspaces=2, not sure how GitHub or other markdown renders handle the legibility issues

justjake:
For me the primary issue is just having a .prettierrc here so when I save source code files in my editor, I don’t produce a big diff

blackmad (approved):

PR #87, opened 2021-05-21
fix(examples): add empty line at eof to fix prettier

Sc0ra:
As i was installing the sdk locally, i came accross this error after running `npm install`
Fixing this style issue fixed it for me


blackmad:
this was fixed in main, thanks for pointing it out - we're putting in better pre-commit checks asap!

Issue #38, opened 2021-05-13
Isomorphic SDK support


pi0:
Hey! Congratulations on public notion API and awesome work on minimal and typed js SDK 💯 

As of the current version, this package seems to only work with node.js, which makes missing lots of opportunities for supporting non-nodejs environments such as Deno, Cloudflare workers, universal frameworks like [Nuxt.js](https://nuxtjs.org/) , and frontend libraries like [vue-notion](https://github.com/janniks/vue-notion)  this can open LOTS of new opportunities for using sdk.

For this change, we need to avoid depending on `got` and using a universal alternative (`fetch`) API. Since `request` is already refactored, it would be easy to allow providing custom client implementation or creating multi-target bundles (see [ohmyfetch](https://github.com/unjs/ohmyfetch) as an example) that importing `@notionhq/client` be as is now but having `@notionhq/client/isomorphic` import (or `/node` import) that is preconfigured with either [node-fetch](https://www.npmjs.com/package/node-fetch) or got or faster node clients like [nodejs/undici](https://github.com/nodejs/undici) / [undici-fetch](https://www.npmjs.com/package/undici-fetch)

I can propose a PR if above seems reasonable for purpose of notion-sdk.

taylorjdawson:
@pi0 I agree that is an important feature, however, I am also a **huge** fan of Sindre Sorhus's work so I would advocate for either:
- a) Just switching for got to [ky](https://github.com/sindresorhus/ky)
- b) Allowing the user to pass in their preferred npm module for requests
```js
import ky from 'ky'
const notion = new Client({
  requestClient: ky,
  auth: process.env.NOTION_TOKEN,
});
```

b-zurg:
It looks like ky allows providing one's own fetch function compatible with the browser API. 

I don't feel like ky provides a huge benefit over fetch - it's not so hard to write one's own wrappers around the API if need be and less abstractions usually means easier debugging in my experience. 

I think passing in a fetch function is a good pattern though.

aoberoi:
thanks for the kind words and for kicking off the discussion. it sounds like there's a few ways we could go about this.

[...]

PR #94, opened 2021-05-22
Rework error types, remove got in favor of node-fetch

justjake:
## What

- Rework error types. This is a breaking change.
- replace `got` with `node-fetch`, similar to whatwg-fetch.
- Allow pluggable `fetch` function that supports the subset of fetch options compatible with `node-fetch` & browser fetch
- use package.json version in user-agent header

## Why

- Partially addresses #60, although does not provide a solution for security or CORS.
- Because we no longer import big expensive dependencies, consumers will be able to subclass `Client` to override the `request` method to implement proxying requests via the server.

## Remarks

- This adds some new code, but also removes some got-specific code of about the same length. The trade-off seems worth it.
- I haven't tested this code yet, and unfortunately we don't have any tests for `Client.prototype.request`. Being able to inject our own `fetch` implementation should make writing tests easier
- I plan to integration test the branch by integrating it into my project https://github.com/justjake/notrition in the coming days, but don't expect rapid movement here -- I'm doing this work as a hobbyist who wants isomorphic fetch, not as a Notion employee.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -48,7 +48,10 @@ import {
   search,
 } from "./api-endpoints"
 import crossFetch from "cross-fetch"
-import { version as PACKAGE_VERSION } from "../package.json"
+import {
+  version as PACKAGE_VERSION,
+  name as PACKAGE_NAME,
+} from "../package.json"
 import { CrossResponse, SupportedFetch } from "./fetch-types"
 
 export interface ClientOptions {
@@ -87,7 +90,7 @@ export default class Client {
   public constructor(options?: ClientOptions) {
     this.#auth = options?.auth
     this.#logLevel = options?.logLevel ?? LogLevel.WARN
-    this.#logger = options?.logger ?? makeConsoleLogger(this.constructor.name)
+    this.#logger = options?.logger ?? makeConsoleLogger(PACKAGE_NAME)
     this.#prefixUrl = (options?.baseUrl ?? "https://api.notion.com") + "/v1/"
     this.#timeoutMs = options?.timeoutMs ?? 60_000
     this.#notionVersion = options?.notionVersion ?? Client.defaultNotionVersion
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -1,5 +1,4 @@
 import type { Agent } from "http"
-import { URL } from "url"
 import {
   Logger,
   LogLevel,
@@ -117,14 +116,16 @@ export default class Client {
 
     // If the body is empty, don't send the body in the HTTP request
     const bodyAsJsonString =
-      body !== undefined && Object.entries(body).length === 0
+      !body || Object.entries(body).length === 0
         ? undefined
         : JSON.stringify(body)
 
     const url = new URL(`${this.#prefixUrl}${path}`)
     if (query) {
       for (const [key, value] of Object.entries(query)) {
-        url.searchParams.append(key, String(value))
+        if (value !== undefined) {
+          url.searchParams.append(key, String(value))
+        }
       }
     }
 
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -397,9 +397,6 @@ export default class Client {
   }
 }
 
-const u = new URL("https://example.com")
-u.searchParams.append
-
 /*
  * Type aliases to support the generic request interface.
  */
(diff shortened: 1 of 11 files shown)


pi0:
Hi, @justjake thanks for the initiative on PR. Regarding #38 and choosing `cross-fetch`, there is a downside that we are still bound to NodeJS specific conventions since the way it works is depending on `browser` and `react-native` fields picked by bundlers (webpack, rollup) and internally using CommonJS syntax. This means still Deno support is impossible and in the browser, `module` script support can be tricky. 

I will propose an alternative approach later today to try.

wesleycoder:
I was trying to use the sdk within Deno for fun and saw that got doesn't work there.
I would advise in implementing `cross-fetch` as a polyfill instead of importing fetch directly from it so this lib could work outside Node environment like on Deno.

Does this makes sense for you? Because I can see some demand of using `@notion/client` within the Deno community.

src/Client.ts
```diff
@@ -43,22 +48,20 @@ import {
...
   search,
 } from "./api-endpoints"
-
-import got, {
-  Got,
-  Options as GotOptions,
-  Headers as GotHeaders,
-  Agents as GotAgents,
-} from "got"
+import crossFetch from "cross-fetch"
```
wesleycoder, line 51: I suggest importing `cross-fetch` as a polyfill `import 'cross-fetch/polyfill'` instead of importing as an instance, then getting it from the global scope because it will be the environment `fetch` if it exists or be the crossFetch polyfill if it doesn't.

This way it may have better chance to work on non Node environments like Deno.

justjake: I don’t think a library should force a specific polyfill on library consumers if it can be avoided.

wesleycoder:
@pi0 if we could get this PR working we can import this package via https://esm.sh/ as it does the heavy lifting on converting the imports to something that Deno can work with.
I dunno about other incompatibilities though.

justjake:
I've tested these changes with my hobby project locally, and they seem legit. My usage: https://github.com/justjake/notrition/pull/2

justjake:
Update: at request of @pllnk I've removed cross-fetch. We don't want to encourage usage of the library in front-end code, because the API's access tokens are unsafe to use from the browser.

src/errors.ts
```diff
@@ -1,132 +1,266 @@
...
+  return isObject(error) && error instanceof NotionClientErrorBase
+}
+
+/**
+ * Narrows down the types of a NotionClientError.
+ * @param error any value, usually a caught error.
+ * @param codes an object mapping from possible error codes to `true`
+ * @returns `true` if error is a `NotionClientError` with a code in `codes`.
+ */
+export function isNotionClientErrorWithCode<Code extends NotionErrorCode>(
```
pllnk, line 79: I think we should only export `isNotionClientError` for now. The error type can be narrowed with equality checks and passing in an object map is cumbersome.

```suggestion
function isNotionClientErrorWithCode<Code extends NotionErrorCode>(
```

pllnk (commented):
Thanks! Could you also:
- List the breaking changes this PR makes and how to resolve them so we can add that to the release notes
- Update the documentation in the README

justjake:
# Release notes

## New features

- The `got` library is replaced with `node-fetch`.
- You may now pass a `fetch` function when creating a `Client` instance. The client will use this fetch function for all HTTP requests.
- Error types have been reworked for improved type safety. No more stringly-typed errors! Inside a `isNotionClientError` type guard, both `error.name` and `error.code` are enumerated unions that can be used to further refine the type of `error`. 
- `error.code` is guaranteed to be in your library's `APIErrorCode` or ClientErrorCode` enum types, even if the API returns a new, unknown `code` in a response. In this case, `error.code` will be `ClientErrorCode.ResponseError`.

## Breaking changes

- The `HTTPResponseError` class is no longer exported, and `HTTPResponseError.isHTTPResponseError(error)` is replaced by `isHTTPResponseError(error)`.
- Most error class constructor signatures have changed. Please see the types for more details. Error class constructors no longer reference the `got` library

pllnk (approved):
Thanks!

huw:
Yo—I noticed the `fetch` field [isn’t documented in the README](https://github.com/makenotion/notion-sdk-js#client-options). Is this intentional (is it supposed to be kind of hidden?)

justjake:
@huw [not intentional](https://en.m.wikipedia.org/wiki/Hanlon's_razor)

huw:
Hahaha good to hear-thanks for supporting this. Gonna build something in Cloudflare Workers!

PR #95, opened 2021-05-23
fix: update image links

maharshmellow:
This PR fixes the broken readme banner images. The rest of the `notion.com` links get redirected to `notion.so` but `dev.notion.com` doesn't so these image urls still need the `.so` tld.


blackmad:
Thanks so much!

PR #101, opened 2021-05-25
Fix CI build failures

seratch:
This pull request fixes the build failure with the latest revision. The command `npm i` fails as it tries to validate the markdown files under `.github` directory.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 build
+.github/**/*.md
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "prepare": "npm run lint && npm run build",
     "build": "tsc",
     "prettier": "prettier --write .",
-    "lint": "prettier --check '**/*.ts' && eslint . --ext .ts && cspell '**/*' ",
+    "lint": "prettier --check . && eslint . --ext .ts && cspell '**/*' ",
     "test": "ava",
     "check-links": "git ls-files | grep md$ | xargs -n 1 markdown-link-check",
     "prebuild": "npm run clean",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "prepare": "npm run lint && npm run build",
     "build": "tsc",
     "prettier": "prettier --write .",
-    "lint": "prettier --check . && eslint . --ext .ts && cspell '**/*' ",
+    "lint": "prettier --check '**/*.ts' && eslint . --ext .ts && cspell '**/*' ",
     "test": "ava",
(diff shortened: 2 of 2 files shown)


justjake:
Point well-taken, but it would be better to exclude only `.github` rather than linting only .ts files.

justjake (approved):

seratch:
@justjake Thanks for checking this! 

I was assuming this project unintentionally applies prettier to any files. If not, another good way to fix this would be to resolve the specific warning about `.github/ISSUE_TEMPLATE/bug_report.md: https://github.com/makenotion/notion-sdk-js/runs/2659621302 That said, this is your project's decision. 

For what it's worth, the following is an alternative way to fix the error, without adding any exclusion.

```
$ npx prettier --check .
Checking formatting...
[warn] .github/ISSUE_TEMPLATE/bug_report.md
[warn] Code style issues found in the above file(s). Forgot to run Prettier?

$ npx prettier -w .github/ISSUE_TEMPLATE/bug_report.md
.github/ISSUE_TEMPLATE/bug_report.md 43ms

$ git diff
diff --git a/.github/ISSUE_TEMPLATE/bug_report.md b/.github/ISSUE_TEMPLATE/bug_report.md
index 3e2cc92..7933254 100644
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,10 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: ''
-labels: ''
-assignees: ''
-
+title: ""
+labels: ""
+assignees: ""
 ---

 Report bugs here only for the Node JavaScript library.
@@ -15,7 +14,7 @@ If you're having problems using Notion's API, or have any other feedback about t
 A clear and concise description of what the bug is.

 **To Reproduce**
-Node version:
+Node version:
 Notion JS library version:

 Steps to reproduce the behavior:

$ npx prettier --check .
Checking formatting...
All matched files use Prettier code style!
```

justjake:
You’re right, but I was confused by the fact that my branch was failing with this error, even though my branch didn’t have the failing markdown file committed. That meant that I couldn’t actually run the fix as you describe. It was weird. Probably some artifact of how GitHub runs pull request CI.

justjake:
We should fix HEAD as you describe. Happy to stamp and merge a PR that does so.

seratch:
@justjake Thanks for your reply! Indeed, your situation seems to be weird. 

I just sent another pull request fixing this: https://github.com/makenotion/notion-sdk-js/pull/110 If it's helpful to this project, please consider merging it.

PR #102, opened 2021-05-25
Remove duplicating type

KentoMoriwaki:
I found `MultiSelectProperty` interface is declared twice, so I deleted one.


lao9 (approved):
Thank you for the cleanup! 🙏

PR #103, opened 2021-05-25
Add typing hints for page update (WIP)

Sc0ra:
Hello !

Since I started using Notion API and this SDK, I wanted to work around typing.
Because the notion API format is quite heavy, it's easy to do mistake when filling properties, at page update for example.

This PoC shows how we could use generic typing to validate the page update method inputs based on the page definition.

Would be glad if you can give me some feedbacks !

Cheers

### DIFF
--- a/examples/typed-page-update/README.md
+++ b/examples/typed-page-update/README.md
@@ -1,0 +1,34 @@
+# Sample Integration: Types Page Update
+
+## About the Integration
+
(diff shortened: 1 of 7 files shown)


justjake:
@Sc0ra We've since switched a new typing system with different types for requests and responses. Types are now more accurate as they are automatically generated from our API.

PR #104, opened 2021-05-25
Fix Annotations being an array and not just an object

thedaviddias:
Currently, the Notion API returns `Annotations` as an array and not just an object.


thedaviddias:
@blackmad can I have someone reviewing this PR? Thks!

lao9:
Hi @thedaviddias - "annotations" is not an array, it is an object. See documentation https://developers.notion.com/reference/rich-text#all-rich-text 

Sample:
```
                "text": [
                    {
                        "type": "text",
                        "text": {
                            "content": "Lacinato kale",
                            "link": null
                        },
                        "annotations": {
                            "bold": false,
                            "italic": false,
                            "strikethrough": false,
                            "underline": false,
                            "code": false,
                            "color": "default"
                        },
                        "plain_text": "Lacinato kale",
                        "href": null
                    }
                ]
            }
```

Please close this PR if you agree!

thedaviddias:
Thanks @lao9, my mistake, I found out where my issue was coming from. Closing it!

PR #105, opened 2021-05-26
Remove duplicate MultiSelectProperty interface

samueldg:
The interface is already defined right above.


samueldg:
Oops! Just noticed there's an identical PR already https://github.com/makenotion/notion-sdk-js/pull/102/files

PR #108, opened 2021-05-28
Changes Block children prop type to Block[]

danecando:
Block children are currently of type BlockBase[]. In my use case I am unable to fetch the children of a particular block and add them to this property because of its type.

Also added .github to prettierignore since it was breaking the checks.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 build
+.github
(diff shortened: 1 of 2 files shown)


lao9 (approved):
Thank you for fixing! 🙏

danecando:
No worries. So just want to confirm I’m correct that property is there to
add the children to as you fetch them recursively? What it seemed like to
me. Thanks!

lao9:
I think your fix is correct that when writing new blocks, you can add nested children using that `children` array.

However, when fetching blocks (using `blocks.children.list()`) we don't return the `children` array, instead we return `has_children` boolean.

I think the types here for read vs. write for block need to be reworked a little bit, but not super related to your change @danecando .

PR #115, opened 2021-06-05
Type predictions for union types

### LINKED ISSUE #92: Can't index a property value by its type when using Typescript
It'd be nice to be able to access the "type-specific data" in a property object without caring about what the type actually is. For example, to be able to do this: `const propertyData = property[property.type]`. Typescript won't let you do this though because there is no index signature on the `PropertyValue` type. 

If I try to write this function:
```
const getAllPropertyData = async () => {
    const databaseId = process.env.NOTION_DATABASE_ID ?? "";
    try {
        const { results } = await notion.databases.query({ database_id: databaseId });
        const propertiesData = results.map(({ properties }) => {
            // type PropertyValue
            const property = properties.Day;

            // "number" | "title" | "rich_text" | "select" ... etc.
            const type = property.type;

            // Type error here.
            const propertyData = property[type];

            return propertyData;
        });
        return propertiesData;
    } catch (error) {
        console.error(error);
    }
}
```
I'll get the following type error:
`Element implicitly has an 'any' type because expression of type '"number" | "title" | "rich_text" | "select" | "multi_select" | "date" | "formula" | "rollup" | "people" | "files" | "checkbox" | "url" | "email" | "phone_number" | "created_time" | "created_by" | "last_edited_time" | "last_edited_by"' can't be used to index type 'PropertyValue'.`

But the `type` always refers to another key in the property object as it states in the [working with databases guide here](https://developers.notion.com/docs/working-with-databases). It makes sense then to have an index signature to allow you to access property data in this way. 

I think something like this would work:
```
export interface PropertyValueBase {
    [typeSpecificDataKey: string]: RichText[] | number | Select // ... etc. ;
    id: string;
    type: string;
}
```

EnixCoda:
This PR provides prediction functions that distinguish union types, which are frequently used in request return values. It works very well in my own use cases.

Usage example (modified `examples/database-update-send-email`):
image

Should help resolve #92 

PS: The added file would look better if `printWidth` of prettier could be set to larger number than the default `80` :P

### DIFF
--- a/src/api-types-prediction.ts
+++ b/src/api-types-prediction.ts
@@ -1,0 +1,347 @@
+import * as APITypes from "./api-types"
+
+function createPrediction<BaseType, Args extends unknown[]>(
+  subTypePredicateCreator: (
+    ...subTypePredictionArgs: Args
+  ) => (theType?: BaseType) => boolean
+) {
+  return function createTypePrediction<SubType extends BaseType>(
+    ...subTypeArgs: Args
+  ) {
+    const subTypePredicate = subTypePredicateCreator(...subTypeArgs)
+    return function isSubType(theType?: BaseType): theType is SubType {
+      return subTypePredicate(theType)
+    }
+  }
+}
+
+/**
+ * Usage:
+ * const isSubType = createPredictionByFunction<BaseType>()<SubType>(theType => theType !== "list")
+ */
+const createPredictionByFunction = <BaseType>() =>
+  createPrediction<BaseType, [(theType?: BaseType) => boolean]>(fn => fn)
+
+/**
+ * Usage:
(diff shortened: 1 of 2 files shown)


justjake:
This is a nice idea, but closing as we're planning to replace many of the existing hand-authored types with OpenAPI-generated types.

EnixCoda:
So you will deprecate the union types since generating the types from OpenAPI? If not, I am very curious how would you resolve union types. Looking forward to your progress.

justjake:
@EnixCoda The types may change a bit as we do the transition, and using an OpenAPI schema may allow us to generate type guard functions as well. For those reasons it doesn’t make sense to accept this changeset at the moment. We can revisit after the new types land.

PR #127, opened 2021-06-24
chore(ci): nodejs v16

liuderchi:
Given that nodejs v16 is the latest LTS, I think we could include it in CI tests.

https://nodejs.org/en/about/releases/


justjake (approved):

justjake:
Thanks for the contribution!

Psoffie009:
> Given that nodejs v16 is the latest LTS, I think we could include it in CI tests.
> 
> https://nodejs.org/en/about/releases/

Issue #131, opened 2021-06-27
Create dynamically options of SelectOption and MultiSelectOption Typescript Bug


usagrada:
Report bugs here only for the Node JavaScript library.

If you're having problems using Notion's API, or have any other feedback about the API, please reach out to support by going to [developers.notion.com](https://developers.notion.com), scroll to the bottom and click on `Help`.

**Describe the bug**
We cannot dynamically create new options for Select or Multi-Select properties with TypeScript.
According to [this page](https://developers.notion.com/changelog/select-values-can-now-be-dynamically-created-via-create-and-update-page-endpoints-other-updates-since-public-beta-launch), it is able to create  new options for Select or Multi-Select properties, but type error is occurred.

**To Reproduce**
Node version: v16.1.0,
Notion JS library version: ^0.2.0
TypeScript version: ^4.3.4
ts-node version: ^10.0.0"

Create Table by slash command in Notion and execute the next commands.

```ts
// code.ts
import { Client } from "@notionhq/client";
import dotenv from "dotenv";
dotenv.config();

const notion = new Client({ auth: process.env.NOTION_ACCESS_TOKEN });
const databaseId = "your database id";
(async () => {
  const response = await notion.pages.create({
    parent: {
      database_id: databaseId,
    },
    properties: {
      Tags: {
        type: "multi_select",
        multi_select: [
          {
            name: "new tag",
            color: "purple",
          },
        ],
      },
      Name: {
        id: "title",
        type: "title",
        title: [
          {
            type: "text",
            text: { content: "new title" },
          },
        ],
      },
    },
  });
})();

```

**Expected behavior**
The upper code will be successfully done.

**Suggestion**
The id of the interface of SelectOption should be optional. (Maybe color is also optional, because the created color is not compatible with the option. However, I think the problem of color is not the problem of this library.)

```ts
export interface SelectOption {
    name: string;
    id: string;
    color: Color;
}
```

**Screenshots**
Please include any screenshots that help explain your problem.

<img width="975" alt="the problem of type" src="https://user-images.githubusercontent.com/61061134/123539009-8145a200-d772-11eb-81fb-46957be7403d.png">

**Additional context**

PR #135, opened 2021-06-28
Require one of name or id in SelectOption, make color optional

### LINKED ISSUE #129: SelectOption and MultiSelectOption Typescript Bug
`SelectOption` and `MultiSelectOption` currently require both `name` and `id`. They should only require one or the other, not both. From the [docs:](https://developers.notion.com/reference/page#multi-select-option-values)

> When updating a multi-select property, you can use either `name` or `id`.	

```
export interface SelectOption {
    name: string;
    id: string;
    color: Color;
}
export declare type MultiSelectOption = SelectOption;
```

(`color` should also probably be marked as optional)

andrewshen:
Changes `api-types.ts` to require one of `name` or `id` in `SelectOption` and make `color` optional.

Fixes #129 and #131.

### DIFF
--- a/src/api-types.ts
+++ b/src/api-types.ts
@@ -336,11 +336,19 @@ export interface BotUser extends UserBase {
  * Misc (output)
  */
 
-export interface SelectOption {
-  name: string
-  id: string
-  color: Color
+export interface SelectOptionBase {
+  color?: Color
 }
+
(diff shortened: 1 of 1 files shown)


justjake (approved):

justjake:
Thanks for the contribution!

PR #142, opened 2021-07-09
Update issue template to use support email

alicial:


justjake (approved):
suggest changing line 10 to `Notion's NodeJS SDK`?

lgtm

PR #146, opened 2021-07-10
Add archived parameter

### LINKED ISSUE #68: How do I delete a database item?
First of all, I want to thank everyone at notion for the API 👏🏼 🥳 

My question is how would I go about deleting a database item using the JS SDK? 

I've tried the following but no go:
```
await notion.request({
    path: `pages/${data.pageId}`,
    method: 'DELETE',
    body: {
        parent: { database_id: databaseId }
    }
})
```

alicial:
Add missing archived parameter, so pages can be deleted.

Fixes #68

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -220,21 +220,25 @@ interface PagesUpdatePathParameters {
 }
 interface PagesUpdateQueryParameters {}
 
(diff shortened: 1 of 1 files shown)


kiranpandit (approved):
👍

PR #150, opened 2021-07-14
Add create database endpoint

kiranpandit:
Adds create database endpoint to Notion API.

Tested by creating database with the client locally.


pllnk (approved):
Thanks!

PR #159, opened 2021-07-22
Add name to property schema object

kiranpandit:
Adding name to property schema object for this [update](https://developers.notion.com/changelog/database-property-objects-now-include-property-name).

### DIFF
--- a/src/api-types.ts
+++ b/src/api-types.ts
@@ -170,6 +170,7 @@ export type Property =
 export interface PropertyBase {
   id: string
   type: string
+  name: string
 }
 
 export interface TitleProperty extends PropertyBase {


pllnk (commented):
Thanks! I think we need to update other types too because this makes `name` required in requests.

kiranpandit:
@pllnk what do you mean required in requests?
`PropertyBase` is used in [Property](https://github.com/makenotion/notion-sdk-js/blob/4811717ace6b6860cd213977bbf620b2839bbdfb/src/api-types.ts#L149) which is part of the response.

pllnk (approved):
Ah there's `PropertyBase` and `PropertyValueBase`. Thanks for clarifying!

PR #164, opened 2021-08-02
Add update and retrieve block endpoints to sdk

almawang:
Adds support for
* PATCH /blocks/:id endpoint
* GET /blocks/:id endpoint
* update helper function to support Union types

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -26,12 +26,68 @@ import {
   ParentPageInput,
   PropertySchema,
   RichTextInput,
+  UpdateBlock,
 } from "./api-types"
 
 // TODO: type assertions to verify that each interface is synchronized to the list of keys in the runtime value below.
 
 // TODO: instead of importing interfaces like BlockBase, should i use a type alias to Block?
 // TODO: need an input version of Block
+
+/*
+ * blocks.retrieve()
+ */
+
+interface BlocksRetrievePathParameters {
+  block_id: string
+}
+interface BlocksRetrieveQueryParameters {}
+interface BlocksRetrieveBodyParameters {}
+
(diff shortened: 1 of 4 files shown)


src/helpers.ts
```diff
@@ -9,7 +9,9 @@ export function assertNever(value: never): never {
   throw new Error(`Unexpected value should never occur: ${value}`)
 }
 
-export function pick<O extends unknown, K extends keyof O>(
+type AllKeys<T> = T extends unknown ? keyof T : never
```
almawang, line 12: this supports getting all the keys from a union type, which occurs in the UpdateBlock request params

src/api-endpoints.ts
```diff
@@ -26,13 +26,69 @@ import {
...
+
+interface BlocksUpdatePathParameters {
+  block_id: string
+}
+interface BlocksUpdateQueryParameters {}
+type BlocksUpdateBodyParameters = UpdateBlock
+
+export type BlocksUpdateParameters = BlocksUpdatePathParameters &
+  BlocksUpdateQueryParameters &
+  BlocksUpdateBodyParameters
```
kiranpandit, line 73: why are these intersected?

kiranpandit (approved):
just one clarifying question, but looks good! 👍

src/api-endpoints.ts
almawang: The `Parameters` for all the other endpoints are interfaces that extend each other so that the final type is the key/value pairs of all the params

For update block this doesn't work since the body param type uses `|`, which isn't supported for interfaces so I used `&` with types instead

kiranpandit (approved):
🎉

PR #168, opened 2021-08-06
Versioned response type change for appendBlockChildren

almawang:


src/Client.ts
```diff
@@ -94,7 +94,7 @@ export default class Client {
   #agent: Agent | undefined
   #userAgent: string
 
-  static readonly defaultNotionVersion = "2021-05-13"
+  static readonly defaultNotionVersion = "2021-07-27"
```
nishadagrawal, line 97: should we just go right to "2021-08-02"?
I don't have any SDK changes to make

nishadagrawal (approved):

nishadagrawal (approved):

PR #170, opened 2021-08-12
Add update database to client

kiranpandit:
Adding support for update database: https://developers.notion.com/reference/update-a-database

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -268,7 +268,7 @@ export interface DatabasesCreateParameters
   extends DatabasesCreatePathParameters,
     DatabasesCreateQueryParameters,
     DatabasesCreateBodyParameters {}
-export interface DatabasesCreateResponse extends BlockBase {}
+export interface DatabasesCreateResponse extends Database {}
 
 export const databasesCreate = {
   method: "post",
@@ -299,7 +299,7 @@ export interface DatabasesUpdateParameters
   extends DatabasesUpdatePathParameters,
     DatabasesUpdateQueryParameters,
     DatabasesUpdateBodyParameters {}
-export interface DatabasesUpdateResponse extends BlockBase {}
+export interface DatabasesUpdateResponse extends Database {}
 
 export const databasesUpdate = {
   method: "patch",
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -27,6 +27,7 @@ import {
   PropertySchema,
   RichTextInput,
   UpdateBlock,
+  UpdatePropertySchema,
 } from "./api-types"
 
 // TODO: type assertions to verify that each interface is synchronized to the list of keys in the runtime value below.
@@ -255,7 +256,7 @@ interface DatabasesCreatePathParameters {}
 interface DatabasesCreateQueryParameters {}
 
 export type InputPropertySchemaMap = {
-  [propertyName: string]: PropertySchema
+  [propertyName: string]: UpdatePropertySchemaMap
 }
 interface DatabasesCreateBodyParameters {
   parent: ParentPageInput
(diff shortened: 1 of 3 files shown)


src/api-endpoints.ts
```diff
@@ -277,6 +278,37 @@ export const databasesCreate = {
...
+interface DatabasesUpdateBodyParameters {
+  properties?: UpdatePropertySchemaMap
+  title?: RichTextInput[]
+}
+
+export interface DatabasesUpdateParameters
+  extends DatabasesUpdatePathParameters,
+    DatabasesUpdateQueryParameters,
+    DatabasesUpdateBodyParameters {}
+export interface DatabasesUpdateResponse extends BlockBase {}
```
almawang, line 302: This returns a `Database` object right? not `BlockBase`?

src/api-types.ts
```diff
@@ -1040,3 +1040,28 @@ export interface LastEditedTimePropertySchema {
 export interface LastEditedByPropertySchema {
```
almawang, line 1040: This is not part of the PR, but why is this a `PropertySchema`? this shouldn't be editable right?

src/api-types.ts
```diff
@@ -1040,3 +1040,28 @@ export interface LastEditedTimePropertySchema {
...
+
+type UpdateSelectOptionSchema = SelectOptionSchema | SelectOption
+
+type UpdateMultiSelectOptionSchema = UpdateSelectOptionSchema
+export interface UpdateSelectPropertySchema {
+  select: { options?: UpdateSelectOptionSchema[] }
+}
+
+export interface UpdateMultiSelectPropertySchema {
+  multi_select: { options?: UpdateMultiSelectOptionSchema[] }
```
almawang, line 1059: Why have a separate type for the mulitselect options?

src/api-types.ts
almawang: Same for `  | CreatedTimePropertySchema
  | CreatedByPropertySchema
  | LastEditedTimePropertySchema
  | LastEditedByPropertySchema`

almawang (changes requested):
I think the `Response` type needs to be updated and remove the non updatable property schema types

src/api-types.ts
kiranpandit: these can be added to the database schema as properties, the values themselves are not updated

src/api-types.ts
kiranpandit: good point - updated so both use the same type

src/api-endpoints.ts
kiranpandit: good catch, fixed!

src/api-types.ts
almawang: ah i see, sg!

src/api-types.ts
```diff
@@ -1040,3 +1040,27 @@ export interface LastEditedTimePropertySchema {
...
+ * Update database property schema (input)
+ */
+export interface RenamePropertySchema {
+  name: string
+}
+
+type UpdateSelectOptionSchema = SelectOptionSchema | SelectOption
+
+export interface UpdateSelectPropertySchema {
+  select: { options?: UpdateSelectOptionSchema[] }
```
almawang, line 1054: the `options` field is nullable? what's the behavior when an empty object is passed into `select`?

almawang (approved):
LGTM, question about the behavior of an empty `select` field

src/api-types.ts
kiranpandit: if it's undefined, then there will be no options (similar to when you create a "select" property in the client before adding any options). In [create database](https://developers.notion.com/reference/create-a-database#select-configuration), the `options` are optional (ha).

PR #173, opened 2021-08-16
fix: NumberPropertySchema type

75asa:
# Summary

There was a difference between the successful body format in Postman and the sdk interface when trying to update the number type in page prop.

# What's change ?

We are updating the interface according to the actual successful body in Postman.

### update number type prop only number

This is exactly what [the documentation](https://developers.notion.com/reference/page#number-property-values) says.

CleanShot 2021-08-16 at 12 08 14

```json
{
    "properties": {
        "prefixNumber": {
            "number": 1
        }
    }
}
```

### update number type prop w/ format

But When w/ format, Something is wrong with the SDK.
missing number value ??

CleanShot 2021-08-16 at 12 10 08

```json
{
    "properties": {
        "prefixNumber": {
            "number": 1,
            "format": "number_with_commas"
        }
    }
}
```

# Note

Please let me know if you have any advice.


75asa:
It's my mistake.
It's solved when updating prop value, used by `NumberPropertyValue`.
thanx

PR #187, opened 2021-09-09
Add relation, rollup, and formula property types to create and update database

kiranpandit:


rhart92 (approved):
Not sure I have full context on the change, but types look like they include relevant properties 👍

PR #227, opened 2021-11-10
Fix notion.users.list arg in README.md

toolness:
`notion.users.list()` is called without any arguments, but (the TypeScript typings, at least) indicate that it needs to take one argument, so this uses an empty object.


lao9 (approved):
Thank you!

PR #231, opened 2021-11-15
Add "link_preview" and "template" blocks to sdk

almawang:


lao9 (commented):
This seems like a lot of new lines 🤨

almawang:
hmm i think the template blocks cause a lot of dupe, lemme see if I do referencing for the open API stuff somehow

PR #241, opened 2021-12-07
updates end points with timezone

alpartr:


cwang314 (dismissed):
🙌

cwang314 (approved):
nice, don't forget to update the version in package.json and publish it to npm

PR #248, opened 2021-12-29
fix(Client): Ensure fetch method is uppercase

dvanoni:
I'm using the Notion SDK in a somewhat unusual environment—a [plugin for Zotero](https://github.com/dvanoni/notero). Zotero is built on the Firefox platform and thus provides a browser-like environment.

I ran into an obscure issue where `retrieve` and `create` methods worked just fine, but `update` methods failed with an unknown network error. I eventually determined the issue was that the `method` passed to `fetch()` needed to be uppercase—`PATCH` instead of `patch`.

I'm not sure why this isn't an issue for the lowercase `get` and `post` requests used by the `retrieve` and `create` methods. I'm also not sure why the conversion to uppercase [performed by node-fetch](https://github.com/node-fetch/node-fetch/blob/1493d046bc0944886277b0b82dfdf78a7b9f7799/src/request.js#L59) isn't working. Perhaps the `fetch()` call is bypassing the node-fetch version and using the environment's built-in `fetch()` method?

Whatever the cause may be, this PR contains a simple fix for the issue.

While waiting for this to get merged into the official package, I've [published my own version](https://github.com/dvanoni/notion-sdk-js/releases/tag/v0.4.12-dvanoni) that I'm currently using in my project.


utkuturunc:
using this library in a chrome plugin also results into `patch` being lowercase

following response is recevied

```
<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<hr><center>cloudflare</center>
</body>
</html>
```

dvanoni:
@wcauchois I see you're the most recent person to push to `main`. Any chance we could get a review on this bug fix PR? Thanks!

xoxohorses (approved):
Thanks for this PR @dvanoni! Appreciate your taking the time to make an improvement to our SDK. We agree that this change would improve the flexibility of casing for the methods we input; merging.

Issue #247, opened 2021-12-27
Error: Property 'properties' does not exist on type '{ object: "page"; id: string; }' - v0.4.11


dharshatharan:
**Describe the bug**
When I upgrade from `v0.4.9` to `v0.4.11`, I get type errors in my code that look like this
```
Property 'properties' does not exist on type '{ object: "page"; id: string; }'
```
Seems like the recent type updates in the `api-endpoints.ts` file has new types that look similar to the following that cause the error:
```typescript
{
    ...
        object: "page";
        id: string;
        created_time: string;
        last_edited_time: string;
        archived: boolean;
        url: string;
    } | {
        object: "page";
        id: string;
    }>;
```
and
```typescript
{
    ...
} | {
    object: "block";
    id: string;
}
```

**To Reproduce**
Node version: `v16.13.0`
Notion JS library version: `v0.4.11`

Steps to reproduce the behaviour:
Try running a similar code snippet with `v0.4.11` to make a notion query and parse some data and you will notice `item.properties` throws a `ts(2339)` error. So will `item.icon`.
```typescript
export async function getBookmarks() {
  const { results } = await notion.databases.query({
    database_id: process.env.NOTION_BOOKMARKS_DATABASE ?? "",
    filter: {
      property: "Status",
      select: {
        equals: "Published",
      },
    },
  });
  return results.map((item) => {
    const properties = item.properties;
    return {
      id: item.id,
      emoji: item?.icon?.type === "emoji" ? item.icon.emoji : "",
      createdAt:
        properties.Created.type === "created_time"
          ? properties.Created.created_time
          : "",
      name:
        properties.Name.type === "title"
          ? properties.Name.title[0].plain_text
          : "",
      type: properties.Type.type === "select" ? properties.Type.select : "",
      url: properties.URL.type === "url" ? properties.URL.url : "",
    } as Bookmark;
  });
}
```

**Expected behaviour**
This wasn't the behaviour prior to the upgrade. `v0.4.9` didn't throw these errors. These extra types seem redundant and don't look like they belong there. Why would you have a page or block type that only has an id and no other metadata (created_time, URL, etc)?

Is this something that needs to be fixed? Is there a way to work around this? 

**Screenshots**
![image](https://user-images.githubusercontent.com/43640229/147509145-6abdaada-4910-440d-a0b3-8c64a8896e57.png)
[...]

PR #251, opened 2022-01-09
Added partials to QueryDatabaseResponse

dca123:
This allows partial properties such as icon, archived & properties to be available without having to check for this property. This provides a fix for #247.  This also seems to be an issue that was offshot from #243 . This is why I've used partials to encapsulate the properties and co. This only aims to fix for the `QueryDatabaseResponse` type, but I believe there are other affected types. If this change was in the right direction, I could fix up the rest. 

Since this is a generated file, I'm not sure if this is the correct way.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -52467,7 +52467,7 @@ export type QueryDatabaseResponse =
           }
       object: "list"
       results: Array<
-        | {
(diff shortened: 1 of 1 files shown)


dca123:
https://github.com/makenotion/notion-sdk-js/issues/247#issuecomment-1012602803 sorts this. Since there is no need for this, I'm closing it

PR #252, opened 2022-01-20
Shrink API types from 2.2mb -> 300kb

justjake:
- Shrink API types from 2.2mb -> 300kb.
- Add typings for `caption` on some block types like code blocks.


rhart92 (approved):
👏

PR #264, opened 2022-02-28
Add issue body to list generated by github issues as blocks

ralsuwaidi:
In addition to generating blocks per issue and filling the properties. The issue is now also filled with notion blocks that is converted from markdown from the github issue body


lao9:
Hi there! Thank you for the suggestion, but for our SDK / Glitch examples, we are keeping these lightweight for now.

We may revisit this soon when we want more robust `appendBlock` content examples. 🙏

PR #269, opened 2022-03-03
chore(Client): typed `notionVersion`

2nthony:


justjake:
@2nthony `"some string literal" | string` in Typescript will reduce to just `string`, so I don't see a reason to make this change.

2nthony:
yeah, I think `notionVersion: 'versionString' | 'versionString' | ...` is more reasonable.

2nthony:
in my case is, i have to find out the correct version number on api doc.

so i wondering, why not with type intellisense.

justjake:
I don't think this offers any intellisense improvement, though, because the type is reduced to just `string`. VS Code and the Typescript playground don't offer any completions for me.

image


[Playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAypDGBLAZo+A1CAnAzog9gHZQC8UA5AEwAMNAtNQIwOPlQA+FNllLL5AKFCQoAcQiFsAQwA2mXAWJk4EJKgzY8RDlBzAsiQgHMBA+ET26EKNAC5Y19fK1KoAIm7U+TN1AD0flAAKlIARlDmALZgMhDAiroArkZGEHo4UABumoo4ZhbAUKmSWLL24iWyzglkbr4BwWER+NGx8doAFlIZhPhJKWnthDhAA)

PR #270, opened 2022-03-06
Shrink property filter API types, add block color types

justjake:
- Shrink property filter API types.
- Add block-level color types.
- Bump version to 1.0.2.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -41,11 +41,25 @@ type UserObjectResponse =
       object: "user"
     }
 
-type TextRequest = string
+type StringRequest = string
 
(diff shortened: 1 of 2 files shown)


justjake:
Published.

src/api-endpoints.ts
```diff
@@ -4401,6 +4401,27 @@ type NumberFormat =
 
 type PartialBlockObjectResponse = { object: "block"; id: string }
 
+type ApiColor =
```
rhart92, line 4404: Interesting that ApiColor is included 🤔 Since it seems like it doesn't seem to have come from your lastest PR.

PR #274, opened 2022-03-16
example didn't reflect new database query API

peernohell:
With the new API, filter in database query should have a "or" or "and" (and maybe some other filter) in the first property of filter and it must be an array.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,12 @@ try {
   const myPage = await notion.databases.query({
     database_id: databaseId,
     filter: {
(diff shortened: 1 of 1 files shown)


lao9:
We still support a single property filter in our latest version.

To keep the simple example simple we haven't updated it to include an `and` or `or` clause unnecessarily:

image


THAT SAID - this example has been updated to use `rich_text` instead of `text`, as that has since changed in later versions.

PR #311, opened 2022-06-30
Add `is-inline` property to create, update and retrieve database

kang8:
I find `is_inline` property in [API reference](https://developers.notion.com/reference/database#all-databases:~:text=false-,is_inline,-boolean). Then I send a request to test it and found that it did create a database for inline.

image

Just as I was doing notion API integration recently, I need this feature, so I sent this PR.

Create inline database example:
```bash
curl --location --request POST 'https://api.notion.com/v1/databases/' \
--header 'Content-Type: application/json' \
--header 'Notion-Version: 2022-02-22' \
--header 'Authorization: {token}' \
--data-raw '{
    "parent": {
        "type": "page_id",
        "page_id": "{page_id}"
    },
    "title": [
        {
            "type": "text",
            "text": {
                "content": "Grocery List"
            }
        }
    ],
    "properties": {
        "Name": {
            "title": {}
        },
        "Date": {
            "date": {}
        }
    },
    "is_inline": true
}'
```
Please replace `{token}` and `{page_id}` with your own generated.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -10203,6 +10203,7 @@ export type GetDatabaseResponse =
       last_edited_time: string
       archived: boolean
       url: string
+      is_inline: boolean
     }
 
 export const getDatabase = {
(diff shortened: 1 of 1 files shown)


kang8:
@lao9, (sorry to mention you). Just for this PR, any feedback?

wcauchois:
We should have fixed this in the latest version of the SDK. Thank you!

PR #319, opened 2022-07-13
Export and name many more response objects

wcauchois:
This PR names and exports many more types in `api-endpoints.ts`, focusing on **response** types. Some categories of types that have been exported:

- Top-level objects like page, block, database, and user (in both full and partial forms).
- Rich text objects
- Block types
- User types
- Page property types

(This change stems from a corresponding update to our code generator.)

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -3,43 +3,49 @@
 
 type IdRequest = string | string
 
+export type PersonUserObjectResponse = {
+  type: "person"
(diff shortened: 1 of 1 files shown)


justjake:
HES

justjake:
YES

filipesmedeiros:
I know this is kind of old, but I can't seem to understand why all those types aren't exported. Any specific reason?

PR #321, opened 2022-07-16
Fix console.log output in example code

marceljuenemann:


lao9 (approved):
Thank you for this improvement!

PR #326, opened 2022-07-20
Support for comments API

wcauchois:
Adds `Client.comments.list` and `Client.comments.create`.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -5635,6 +5635,21 @@ export type PropertyItemObjectResponse =
   | RelationPropertyItemObjectResponse
   | RollupPropertyItemObjectResponse
 
+export type CommentObjectResponse = {
+  object: "comment"
+  id: string
+  parent:
+    | { type: "page_id"; page_id: IdRequest }
+    | { type: "block_id"; block_id: IdRequest }
+  discussion_id: string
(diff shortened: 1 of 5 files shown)


src/api-endpoints.ts
```diff
@@ -10358,3 +10373,49 @@ export const search = {
...
+  path: (): string => `comments`,
+} as const
+
+type ListCommentsQueryParameters = {
+  block_id: IdRequest
+  start_cursor?: string
+  page_size?: number
+}
+
+export type ListCommentsParameters = ListCommentsQueryParameters
```
xoxohorses, line 10404: @wcauchois why do we need to create an extra type just to set it equal to an existing type?

xoxohorses (dismissed):
Yay! Looks great. Thanks for working on this :)

src/api-endpoints.ts
wcauchois: this is just part of the codegen, it's all automatic. if we had body parameters as well as query parameters, this type would be `ListCommentsBodyParameters | ListCommentsQueryParameters` - so it makes more sense when it's a union like that.

the codegen isn't special cased to eliminate this alias if the endpoint only has one type of parameters. i don't really mind, it doesn't affect anything too negatively and keeps things consistent with other endpoints!

xoxohorses (approved):

Issue #329, opened 2022-07-25
Paginated utility functions fail for search


Kerrick:
**Describe the bug**
Paginated utility functions fail for search. This includes both `iteratePaginatedAPI` and `collectPaginatedAPI`.

**To Reproduce**

Node version: v16.15.1

Notion JS library version: 2.1.0

Illustrative code that reproduces the behavior:

```ts
import { Client, collectPaginatedAPI, iteratePaginatedAPI } from '@notionhq/client'

export default class Example {
  private client = new Client({ auth: 'REDACTED', notionVersion: '2022-06-28' })

  public async example() {
    for await (const result of iteratePaginatedAPI(this.client.search, {
      filter: { property: 'object', value: 'page' },
    })) {
      console.log(result)
    }
  }
}
```

**Expected behavior**

I would expect it not to throw an exception, and to instead have the behavior shown in the documentation and when using these utilities with other paginated methods. In the above example, each result should be logged to the console. Instead, the following exception is thrown:

```
TypeError: Cannot read properties of undefined (reading 'request')
    at search (/path/to/node_modules/@notionhq/client/src/Client.ts:511:17)
    at iteratePaginatedAPI (/path/to/node_modules/@notionhq/client/src/helpers.ts:50:49)
    at iteratePaginatedAPI.next (<anonymous>)
    at example.example (/path/to/example.ts)
```

lao9:
Thanks so much for spotting - I believe I found a root cause and will have a fix out today.

Kerrick:
> Thanks so much for spotting

Thanks so much for the utility functions 😁

PR #330, opened 2022-07-25
update search to use arrow function

lao9:
Closes https://github.com/makenotion/notion-sdk-js/issues/329

Previous TypeError:
```
TypeError: Cannot read properties of undefined (reading 'request')
    at search (/path/to/node_modules/@notionhq/client/src/Client.ts:511:17)
    at iteratePaginatedAPI (/path/to/node_modules/@notionhq/client/src/helpers.ts:50:49)
    at iteratePaginatedAPI.next (<anonymous>)
    at example.example (/path/to/example.ts)
```

Passing in the `listFn` as `client.search` meant that `this.request({...})` is unavailable unless `search` is an arrow function.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -507,7 +507,9 @@ export default class Client {
   /**
    * Search
    */
-  public search(args: WithAuth<SearchParameters>): Promise<SearchResponse> {
+  public search = (
+    args: WithAuth<SearchParameters>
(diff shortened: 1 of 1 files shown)


Kerrick (approved):

wcauchois (approved):
Great, thanks! Do you think this is worth a 2.1.1 release? 😄

PR #337, opened 2022-08-16
FIx typescript example code

devethan:


xoxohorses (approved):
Thanks for the contribution @devethan !

PR #341, opened 2022-09-01
Support status property, toggle headers, and page property values

s-h-a-r-c:
## Status property
Now supported! [See API changelog](https://developers.notion.com/changelog/changes-for-august-15-28-2022)

## Toggle headers
Now supported! [See API changelog](https://developers.notion.com/changelog/changes-for-august-15-28-2022)

## Page properties
Previously, responses that included page property data would not include page property types or values. This behavior impacted retrieving pages, querying databases, and searching. With this change, page properties will now include types and values in the response, which matches the behavior in version `2022-02-22` of the Notion API.

[See API changelog](https://developers.notion.com/changelog/changes-for-august-31-2022)

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -47,6 +47,751 @@ export type UserObjectResponse =
   | PersonUserObjectResponse
   | BotUserObjectResponse
 
+type StringRequest = string
+
+type SelectColor =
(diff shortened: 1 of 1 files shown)


wcauchois (approved):
:stamping-intensifies:

PR #348, opened 2022-09-29
feat(workspaceName): add workspace_name to bot object

hazelguo:
This PR includes:
- workspace_name change
- `ButtonPropertyItemObjectResponse` from simon
- `comment_id` change from maybe clara

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -4404,7 +4404,6 @@ export type PageObjectResponse = {
       }
     | { type: "last_edited_time"; last_edited_time: string; id: string }
     | { type: "formula"; formula: FormulaPropertyResponse; id: string }
-    | { type: "button"; button: null; id: string }
     | { type: "title"; title: Array<RichTextItemResponse>; id: string }
     | { type: "rich_text"; rich_text: Array<RichTextItemResponse>; id: string }
     | {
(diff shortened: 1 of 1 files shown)


src/api-endpoints.ts
```diff
@@ -4681,6 +4683,13 @@ type LastEditedTimeDatabasePropertyConfigResponse = {
   name: string
 }
 
+type ButtonDatabasePropertyConfigResponse = {
```
cwang314, line 4686: @hazelguo this looks handwritten - if you don't add this, what happens to the types in this file? we shouldn't be exposing this property in the SDK yet (and probably not ever tbh)

src/api-endpoints.ts
```diff
@@ -7379,6 +7406,7 @@ type BlockObjectRequest =
               link_to_page:
                 | { page_id: IdRequest; type?: "page_id" }
                 | { database_id: IdRequest; type?: "database_id" }
+                | { comment_id: IdRequest; type?: "comment_id" }
```
cwang314, line 7409: this is fine - keep these in

cwang314 (changes requested):

src/api-endpoints.ts
hazelguo: I didn't handwrite anything. Everything is from `notion build-public-api-js-sdk --path=../notion-sdk-js/src/`

cwang314 (approved):
thanks for making requested changes!

PR #352, opened 2022-10-04
update api for page create to include parent page_id

hweeks:
the default params for creating a page in the docs indicate a database or parent page are acceptible. as the types stand they only accept a database id and require a cast. this is modeled after the other places that include this, such as the UpdatePageBodyParamters directly under this one.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -9407,7 +9407,7 @@ export const listUsers = {
 
 type CreatePageBodyParameters =
   | {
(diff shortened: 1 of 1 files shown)


lao9 (commented):
If you look here, there's a type specifically for creating a page with a page parent.

It's separated because the only property you can set for a page with a page parent is title.

https://github.com/makenotion/notion-sdk-js/blob/6ce697f29de208b4c6d4cad23d4ab7b28862c93f/src/api-endpoints.ts#L9415-L9429

Issue #345, opened 2022-09-27
Wrong version of Typescript


tgerboui:
**Describe the bug**
In the [requirements](https://github.com/makenotion/notion-sdk-js#requirements) of the README file, it is mentioned that the Typescript version is  >= 4.2. 
However in [fetch-types.ts](https://github.com/makenotion/notion-sdk-js/blob/main/src/fetch-types.ts) file, `Awaited` type is used while it is only [implemented in typescript in 4.5](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#the-awaited-type-and-promise-improvements).
This produces an error when compiling with Typescript < 4.5 :
```bash
node_modules/@notionhq/client/build/src/fetch-types.d.ts:4:30 - error TS2304: Cannot find name 'Awaited'.
```

**To Reproduce**
Node version: 14.17.0
Notion JS library version: 2.2.0

Import something from `@notionhq/client/build/src/Client` with Typescript version 4.2 and notion client 2.2.0.

E.g with an empty project :
```bash
npm init
npm install typescript@4.2
npm install @notionhq/client@2.2.0
echo "
import { ClientOptions } from '@notionhq/client/build/src/Client';
const options: ClientOptions = {};
" > index.ts
./node_modules/.bin/tsc index.ts
```

**Expected behavior**
Have the right Typescript version (>= 4.5) describe in documentation or compile with the specified version (remove `Awaited`)

PR #354, opened 2022-10-06
update typescript and eslint package dependencies

lao9:
closes https://github.com/makenotion/notion-sdk-js/issues/345

Call me dependabot cuz I be depending.

~Updated all dependencies while I was at it, why not?~ Oh right because it's a nightmare.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,14 +40,14 @@
     "build/src/**"
   ],
   "dependencies": {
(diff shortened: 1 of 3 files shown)


wcauchois (approved):
lgtm!

wcauchois (approved):

PR #358, opened 2022-10-10
 add some more type exports

hweeks:
hello again! i'm taking another shot at updating these types, as of today, the request api objects aren't exported, so if you want to get them you gotta do illegal stuff like this:

```ts
import { AppendBlockChildrenParameters } from '@notionhq/client/build/src/api-endpoints';

type Unarray<T> = T extends Array<infer U> ? U : T;

export type DerivedBlockRequest = Unarray<AppendBlockChildrenParameters["children"]>
```

i'd really like to just export these in the same way that `BlockObjectResponse` is

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -5956,7 +5956,7 @@ type RichTextItemRequest =
       }
     }
 
-type BlockObjectRequestWithoutChildren =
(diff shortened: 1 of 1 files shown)


wcauchois:
Hi! This will get fixed with #366, hopefully we'll be able to merge that soon.

Closing for now but thanks for the report!

PR #375, opened 2022-12-19
2.2.3

parulbaweja:
Follow-up from #366 for `filter_properties` query param.

Q: I have created the 2.2.3 tag and pushed it to GitHub, but I don't see it associated with my commit here. Is that expected?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@notionhq/client",
-  "version": "2.2.2",
+  "version": "2.2.3",
   "description": "A simple and easy to use client for the Notion API",
   "engines": {
     "node": ">=12"


parulbaweja:
> Hmm yeah we do need to push the new tag to GitHub before updating `package.json`, but `2.2.3` doesn't seem to appear in the releases page. Were there any steps skipped from [this doc](https://dev.notion.so/notion/Update-Notion-JS-SDK-57cce3061ed04a33b4663fee8f6f9ad1)?

I think we're good now -- I re-ran the command to create the commit + tag. You can see the tag is associated with the commit in this PR [here](https://github.com/makenotion/notion-sdk-js/releases/tag/v2.2.3).

PR #403, opened 2023-05-02
Add peruvian_sol

parulbaweja:
We now support `peruvian_sol`, so updating the SDK. The [API docs](https://developers.notion.com/reference/property-object#number) have been updated as well.


rhart92 (approved):
:shipit:

PR #404, opened 2023-05-02
Upgrade to 2.2.5

parulbaweja:
Includes following changes:
- Fix bug for in `filter_properties` query parameter to correctly URI encode property IDs when passed via the JS SDK
- Add `peruvian_sol` to acceptable `NumberFormat` options


rhart92 (approved):
:shipit:

PR #409, opened 2023-06-11
Adding unique_id to properties of PageObjectResponse

ShotaKarato:
Hi there, 
I just noticed `PageObjectResponse` in `api-endpoints.d.ts` might be missing the definition for `unique_id`??

I have assigned a unique id to each of my page in the DB but when I try to access that id TS compiler throws an error.

Screenshot 2023-06-11 at 16 59 46

> // Error message I got
Property 'unique_id' does not exist on type '{ type: "number"; number: number | null; id: string; } | { type: "url"; url: string | null; id: string; } | { type: "select"; select: SelectPropertyResponse | null; id: string; } | ... 16 more ... | { ...; }'.
  Property 'unique_id' does not exist on type '{ type: "number"; number: number | null; id: string; }'.ts(2339)

But the data I retrieved from the DB includes id, the type of which is `unique_id`.
```TypeScript
{
  id: 'auto-generated-id-comes-here',
  type: 'unique_id',
  unique_id: { prefix: null, number: 'number-comes-here' }
}
```

## Changes I made
- [x] added the definition of `unique_id` based on the result I got from the DB

Any feedback or comments are much appreciated, 
thank you

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -4358,6 +4358,14 @@ export type PageObjectResponse = {
     | { type: "workspace"; workspace: true }
   properties: Record<
     string,
+    | {
+        type: "unique_id";
+        unique_id: {
+            prefix: string | null;
+            number: number;
+        };
+        id: string;
+    }
     | { type: "number"; number: number | null; id: string }
     | { type: "url"; url: string | null; id: string }
     | { type: "select"; select: SelectPropertyResponse | null; id: string }


rhart92:
Hey @ShotaKarato thanks for flagging that `unique_id` was missing from the SDK! As our types are generated, I can't approve your PR but I have updated the SDK from our side. I'm going to close this PR since the change you're suggesting is now a part of v2.2.6 🥳

PR #411, opened 2023-06-27
[Update SDK] Add support for unique ID, public URL, and appending after a block

rhart92:
Add support for:
- Unique ID for pages
- Public URL (if available)
- Appending after a block

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -4400,6 +4400,11 @@ export type PageObjectResponse = {
       }
     | { type: "last_edited_time"; last_edited_time: string; id: string }
     | { type: "formula"; formula: FormulaPropertyResponse; id: string }
+    | {
+        type: "unique_id"
+        unique_id: { prefix: string | null; number: number | null }
+        id: string
+      }
     | { type: "title"; title: Array<RichTextItemResponse>; id: string }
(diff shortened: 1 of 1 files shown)


kiranpandit (approved):
👍

src/api-endpoints.ts
```diff
@@ -10192,7 +10207,10 @@ type AppendBlockChildrenPathParameters = {
   block_id: IdRequest
 }
 
-type AppendBlockChildrenBodyParameters = { children: Array<BlockObjectRequest> }
+type AppendBlockChildrenBodyParameters = {
+  children: Array<BlockObjectRequest>
+  after?: IdRequest
```
dvanoni, line 10212: @rhart92, is there a description of what the `after` parameter is used for? I don't see it mentioned in the [documentation](https://developers.notion.com/reference/patch-block-children), but it sounds intriguing!

rhart92: Hey @dvanoni. The documentation is coming 🔜 but for now if you want to append blocks to somewhere other than then end of the parent block, you can try passing the block ID you want to append after in the `after` prop ⚡ Essentially it just provides a bit more customizability to the [Append block children](https://developers.notion.com/reference/patch-block-children) endpoint.

dvanoni: Thanks for the info! I was hoping that's what it did. Sounds super useful!

PR #416, opened 2023-07-03
Add block type example

jessmitch42:
Adds example of how to parse the text from any type of block that is currently supported by the public API. This example includes retrieving the block children for a page (notion.blocks.children.list()) and parsing the text from each block returned. The text is then printed to the command line.
The page used can be set in the .env file, along with the integration key. 
This example is not intended to have any "real-word" application other than providing devs with some helpful functions to use in their own Notion apps.


examples/parse-text-from-any-block-type/index.js
```diff
@@ -0,0 +1,148 @@
...
+const notion = new Client({ auth: apiKey })
+
+/* 
+---------------------------------------------------------------------------
+*/
+
+// Take rich text array from a block child that supports rich text and return the plain text.
+// Note: All rich text objects include a plain_text field.
+const getPlainTextFromRichText = richText => {
+  return richText.map(t => t.plain_text).join(", ")
```
rhart92, line 18: nit: I think we want to join with just `""` vs. `", "` unless we are trying to illustrate that it's a list? 🤔

examples/parse-text-from-any-block-type/index.js
```diff
@@ -0,0 +1,148 @@
...
+  let source, caption
+
+  if (block[block.type].external) {
+    source = block[block.type].external.url
+  } else if (block[block.type].file) {
+    source = block[block.type].file.url
+  } else if (block[block.type].url) {
+    source = block[block.type].url
+  } else {
+    source = "Missing case for media block types: " + block.type
```
rhart92, line 33: If we aren't going to throw an error here, maybe we should surround with square brackets or something to make it a bit more indicative that it's not user text?

examples/parse-text-from-any-block-type/index.js
```diff
@@ -0,0 +1,148 @@
...
+  if (block[block.type].caption.length) {
+    caption = getPlainTextFromRichText(block[block.type].caption)
+    return caption + ": " + source
+  }
+  // If no caption, just return the source URL
+  return source
+}
+
+// Parse the plain text from any block type supported by the public API.
+const parseText = block => {
```
rhart92, line 45: nit: We aren't really parsing here... maybe `getTextFromBlock`?

examples/parse-text-from-any-block-type/index.js
```diff
@@ -0,0 +1,148 @@
...
+        // Does not include text from ToC; just the color
+        text = "ToC color: " + block.table_of_contents.color
+        break
+      case "breadcrumb":
+      case "column_list":
+      case "divider":
+        text = "No text available"
+        break
+      default:
+        text = "Needs case added"
```
rhart92, line 106: nit: Maybe add the block type here? Should we also add a type for `unsupported`?

examples/parse-text-from-any-block-type/index.js
```diff
@@ -0,0 +1,148 @@
...
+    text = text + " (Has children)"
+  }
+  // Includes block type for readability. Update formatting as needed.
+  return block.type + ": " + text
+}
+
+async function retrieveBlockChildren(id) {
+  console.log("Retrieving blocks (async)...")
+
+  const blocks = await notion.blocks.children.list({
```
rhart92, line 123: Should we include pagination here? We have a helper called `iteratePaginatedAPI` included in the SDK 🤷

rhart92 (dismissed):
Added some comments but looks good 👍

examples/parse-text-from-any-block-type/index.js
jessmitch42: I don't have a strong opinion on this tbh -- updating!

examples/parse-text-from-any-block-type/index.js
jessmitch42: The block type is included in the return at the bottom so all good there. And unsupported is captured above the switch case but I'll move it down to keep all types together.

examples/parse-text-from-any-block-type/index.js
jessmitch42: oh cool, thanks! I'll add it

jessmitch42: Added! Thanks for the review. 🙏 Looks like it needs another look to merge so I re-requested another one.

rhart92 (approved):

PR #428, opened 2023-07-24
Endpoint changes for wikis

kiranpandit:
Changes to support wikis in the public API:
- Query database returns pages and databases.
- Databases can be created as children of databases.
- Pages that are parented by a page can have non title properties.


rhart92 (dismissed):
Looks good! Mind manually removing the OAuth part at the end? I'd like it to be separate both to make sure it can be tested independently + so I can update `Client.ts` with the new method.

rhart92 (approved):

PR #437, opened 2023-08-10
Add webform demo app to /examples

jessmitch42:
This PR adds a new demo app to the examples directory. 

A new draft for the [Create an integration guide](https://developers.notion.com/docs/create-a-notion-integration) has been written and approved, which uses this sample code. (It'll be added to Readme after this PR is merged.)

The goal of this demo is to provide some simple webforms that developers can test out to see how user interface interactions can trigger Notion public API requests. Each form returns an ID that can be used in the next form. (e.g. the "create a database" form returns an ID that can be used in the "create a page" form below it.

An extended version of this type of demo would be a blog site that uses Notion as a CMS.  
Screen Shot 2023-08-10 at 4 01 52 PM


examples/web-form-with-express/public/client.js
```diff
@@ -0,0 +1,136 @@
...
+const blocksResponse = document.getElementById("blocksResponse")
+const commentResponse = document.getElementById("commentResponse")
+
+/**
+ * Functions to handle appending new content to /views/index.html
+ */
+
+// Appends the API response to the UI
+const appendApiResponse = function (res, el) {
+  console.log(res)
```
ronovan, line 24: q -- was this console.log meant to be left in? more than okay if so, just pointing out in case it wasn't

same with other console.log's in this file

examples/web-form-with-express/public/client.js
```diff
@@ -0,0 +1,136 @@
...
+  el.appendChild(result)
+  // See browser console for more information
+  if (res.message === "error") return
+
+  const id = document.createElement("p")
+  id.innerHTML = "ID: " + res.data.id
+  el.appendChild(id)
+
+  if (res.data.url) {
+    var url = document.createElement("a")
```
ronovan, line 37: super nit: can we use `let` instead of `var`?

```suggestion
    let url = document.createElement("a")
```

examples/web-form-with-express/public/client.js
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Functions to handle appending new content to /views/index.html
+ */
+
+// Appends the API response to the UI
+const appendApiResponse = function (res, el) {
+  console.log(res)
+
+  const result = document.createElement("p")
```
ronovan, line 26: optional, nit: it was a bit hard to grok this code because of variables `res` and `result`, separately. Could `result` be renamed to something like `newHtmlElement`, `newParagraph`, `resultElement`, or something along those lines?

if you make this change, would be great to do that in functions below as well

ronovan (approved):
🔥

examples/web-form-with-express/public/client.js
jessmitch42: In this case yes, just so they could see the full response (including an error). A few pieces of data are appended to the UI but I left the console logs in in case it's useful to the person testing it out to see the full response.

examples/web-form-with-express/public/client.js
jessmitch42: good point, yes! Ty!

jessmitch42:
Thanks for the review @ronovan ! Looks like it needs another approval after more commits were added 😇

ronovan (approved):
Love all the new variable names! ✨

Issue #425, opened 2023-07-17
notionhq/client and Next.js 13.4.9 with server action


jercomio:
I'm using notionhq/client with Next.js 13.4.9 and TypeScript for an App project and I have the following issue :

This is  the function to get database :
`export async function getDatabase(databaseId: string) {
    try {
        const response = await notion.databases.query({
            database_id: databaseId!
        })
        return response
    } catch (error: unknown) {
        if (isNotionClientError(error)) {
            switch (error.code) {
                case ClientErrorCode.RequestTimeout:
                    console.log(`RequestTimeout error: ${error.code}`);
                    break
                case APIErrorCode.ObjectNotFound:
                    console.log(`ObjectNotFound error: ${error.code}`);
                    break
                case APIErrorCode.Unauthorized:
                    console.log(`Unauthorized error: ${error.code}`);
                    break
                default:
                    console.log(`Default error: ${error.code}`);
                    break
            }
        }
    }
}`

Then, this is a function using a server action : 
`async function getData() {
    'use server'
    const data = await getDatabase(databaseId!)
    return data
  }`

And I call a client component like that :
`<Component data={getData()} />`

Now, in my component if I write `console.log(data)`, here the response I have : 

`Chunk {
reason: null,
status: "resolved_model",
value: "{...}",
_response: {•••}
}`

To get a good response and use it, I have to write this in useEffect for example :
`console.log(JSON.parse(data.value)`

And now I have my array of object from Notion database and I can use it.

devjiwonchoi:
First of all, please provide an issue in a readable syntax.
You can write codes in ` ```ts <code> ``` ` format and your codes will be much readable.

Since your `getData` is an async function, there for you need to `await` to retrieve data otherwise it will return a `Promise`.

```ts
await getData()
```

Note that in order to use `await`, the Page Component needs to be `async`.

```ts
// page.tsx

export default async function Home() {
  // ...
```

Simple example:

```ts
export default async function Bio() {
  async function getData() {
    'use server'
    return await notionClient.databases.query({
      database_id: process.env.NOTION_DATABASE_ID as string,
    })
  }
  const data = await getData()

  return (<Component data={data} />)
}
```

jercomio:
Yes, Thanks ! 
I updated my code and I get a right result now 👍 
```ts 
// page.tsx

export default async function Home() {

  async function getData() {
    'use server'
    return await getDatabase(databaseId!)
  }
  const data = await getData()

[...]

Issue #443, opened 2023-08-16
Type mismatch for relations. SDK types do not expect to see "has_more"


Matlor:
I fetch data from the notion api. I want to work with that data and have used the types "PageObjectResponse" and "PartialPageObjectResponse" from the sdk to define the responses I can get. For all the properties that are relations I get a type error:

'has_more' does not exist in type '{ type: "relation"; relation: { id: string; }[]; id: string; }'

The data looks like this:

Parent: { id: "%60J%3D~", type: "relation", relation: [{ id: "a8b666cc-4c8c-4e10-b612-2eb1f968b1a4" }], has_more: false, },

devjiwonchoi:
Are you using the method `retrieve`?

narumincho:
Page properties `has_more` is present in the document and the actual response returned

https://github.com/makenotion/notion-sdk-js/blob/7950edc034d3007b0612b80d3f424baef89746d9/src/api-endpoints.ts#L4588

https://developers.notion.com/reference/page-property-values#relation

RaymondWHZ:
I'm facing the same issue here. Hope this can get solved 🙏

PR #445, opened 2023-08-18
fix: modify response types to intersection

devjiwonchoi:
### Description

Since Response types are currently in a union, the users cannot access response objects' property directly.

Example:

```ts
const response = await notion.blocks.children.list({
  block_id: block_id,
})

const archives = response.results.filter((block) => block.archived) // Error here
```

```console
Property 'archived' does not exist on type 'PartialBlockObjectResponse | BlockObjectResponse'.
Property 'archived' does not exist on type 'PartialBlockObjectResponse'.ts(2339)
```

### Fix

Current:

https://github.com/makenotion/notion-sdk-js/blob/83ac0435379c511e3b5a879fcdb06df0e80f5eae/src/api-endpoints.ts#L10279-L10286

Fix:

```ts
// api-endpoints.ts
export type ListBlockChildrenResponse = {
    type: "block";
    block: EmptyObject;
    object: "list";
    next_cursor: string | null;
    has_more: boolean;
    results: Array<PartialBlockObjectResponse & BlockObjectResponse>; // union to intersection
};
```

Fixes  https://github.com/makenotion/notion-sdk-js/issues/425#issuecomment-1683981412 #443


rhart92:

PR #450, opened 2023-08-25
Add support for unique ID in database request / responses

rhart92:
Now `createDatabase` and `updateDatabase` allow creating unique ID properties and `getDatabase` will now return unique ID properties.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -4707,6 +4707,13 @@ type RollupDatabasePropertyConfigResponse = {
   name: string
 }
 
+type UniqueIdDatabasePropertyConfigResponse = {
+  type: "unique_id"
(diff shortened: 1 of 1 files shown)


src/api-endpoints.ts
```diff
@@ -5210,6 +5218,7 @@ type LanguageRequest =
   | "mathematica"
   | "mermaid"
   | "nix"
+  | "notion formula"
```
parulbaweja, line 5221: why was this added?

parulbaweja (approved):

src/api-endpoints.ts
rhart92: Just an artifact of someone adding a new language but not publishing the `sdk` 😢

PR #456, opened 2023-09-24
Update index.js

adasq:
The current version doesn't support a general `rich_text` case, for the following block:

```
{
  type: 'rich_text',
  rich_text: [
    {
      type: 'text',
      text: [Object],
      annotations: [Object],
      plain_text: 'plain!',
      href: null
    }
  ]
}
```

The first if condition:

```
if (block[block.type].rich_text) {
        // This will be an empty string if it's an empty line.
        text = getPlainTextFromRichText(block[block.type].rich_text)
    }
```

doesn't cover it, as it resolves to `block.rich_text.rich_text`.


rhart92:

PR #464, opened 2023-10-11
Use `btoa` when `Buffer` isn’t available

### LINKED ISSUE #444: `Buffer.from` isn’t supported in all JavaScript runtimes
The new `oauth.token` method uses [`Buffer.from` under the hood](https://github.com/makenotion/notion-sdk-js/blob/83ac0435379c511e3b5a879fcdb06df0e80f5eae/src/Client.ts#L188-L189), but `Buffer` is Node-specific and won’t work in other runtimes.

The recommendations for other runtimes appear to be:
- Cloudflare Workers: [`btoa`](https://developers.cloudflare.com/workers/runtime-apis/web-standards/#base64-utility-methods)
- Deno: [`btoa`](https://deno.land/api@v1.36.1?s=btoa) or [`std/encoding/base64`](https://deno.land/std@0.198.0/encoding/base64.ts)
- Vercel Edge Functions: [`btoa`](https://vercel.com/docs/concepts/functions/edge-functions/edge-runtime)
- Web (god forbid): `btoa`

In fact, WinterCG (the standards body governing alternative JS runtime interoperability) [mandates `btoa` on the global scope](https://blog.cloudflare.com/standards-compliant-workers-api/) for compliant runtimes (which is basically any edge runtime that’s not Node or Bun).

The merits of using `btoa` are mixed. [It’s not actually a deprecated API; it’s merely marked at legacy](https://github.com/nodejs/node/issues/40754) and the recommendation for modern code (especially when using Unicode, which we don’t) is to use something more compatible. If you have a single implementation with known inputs and you know it’s going to create the output correctly, it’s probably fine to use `btoa`.

I recommend this package uses `btoa` instead of `Buffer.from().toString(“base64”)`, because it’s more compatible and doesn’t introduce _that much_ of a footgun. The alternative would be to [ponyfill `buffer`](https://github.com/feross/buffer), or introduce a separate package just for base64 encoding/decoding; these would add more complexity and downstream dependencies so are probably less favourable.

huw:
Recreates #461
Fixes #444

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -186,7 +186,10 @@ export default class Client {
       // authorization header as required by `Basic` auth.
       const unencodedCredential = `${auth.client_id}:${auth.client_secret}`
       const encodedCredential =
-        Buffer.from(unencodedCredential).toString("base64")
+        typeof Buffer !== "undefined"
+          ? Buffer.from(unencodedCredential).toString("base64")
+          // Use `btoa` in non-Node environments
+          : btoa(unencodedCredential)
       authorizationHeader = { authorization: `Basic ${encodedCredential}` }
     } else {
       // Otherwise format authorization header as `Bearer` token auth.


cwang314:

huw:
Thanks @cwang314. If I may give some feedback, I'm finding it hard to commit to doing even more open source labour for Notion when this PR has been open for months, contains an extremely simple fix to broken functionality, and hasn't even been acknowledged by your team (aside from previously having to re-open it from a previous git history reset). I don't think I will go to the effort of reopening it.

PR #476, opened 2023-12-04
Update SDK with file block name, toggle table children, more emoji, more timezones

cwang314:
Ran `build-public-api-js-sdk` to pick up the newest API changes


rhart92 (approved):
🙇 Thank you!

PR #479, opened 2023-12-17
feat: Log the response body when the request succeeds with LogLevel.DEBUG

whatasame:
## Issue

The README mentions that the response body will be logged on the DEBUG level, but the actual code currently supports logging only in error response cases. See following excerpt from the [README.md](https://github.com/makenotion/notion-sdk-js#logging)

> If you're debugging an application, and would like the client to log response bodies, set the logLevel option to LogLevel.DEBUG.

## Implementation

Added a log method for success response cases.

> [!NOTE]
> 
> This produces a non-pretty JSON string to prevent unintended creation of strings on non-DEBUG levels.

## Example

image

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -220,6 +220,7 @@ export default class Client {
 
       const responseJson: ResponseBody = JSON.parse(responseText)
       this.log(LogLevel.INFO, `request success`, { method, path })
+      this.log(LogLevel.DEBUG, `succeed response body`, { body: responseText })
       return responseJson
     } catch (error: unknown) {
       if (!isNotionClientError(error)) {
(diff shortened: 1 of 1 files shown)


src/Client.ts
```diff
@@ -220,6 +220,7 @@ export default class Client {
 
       const responseJson: ResponseBody = JSON.parse(responseText)
       this.log(LogLevel.INFO, `request success`, { method, path })
+      this.log(LogLevel.DEBUG, `succeed response body`, { body: responseText })
```
varunrau, line 223: could you make this consistent with the previous log?
```suggestion
      this.log(LogLevel.DEBUG, `response body success`, { body: responseText })
```

whatasame: Hello @varunrau! Thank you for the comment.

I believe your opinion is appropriate, but I have added it to align with the pattern seen in the after log, especially considering the 230-241 lines handling errors.

When an error occurs in the request
 * request fail -> failed response body

When the request is succeeded
 * request success -> **succeeded response body** (what I added)

cwang314:

PR #490, opened 2024-02-18
fix: wrong types in isMentionRichTextItemResponse

alfonsusac:
The return type of the isMentionRighTextitemResponse is supposed to be the corresponding RichTextItemResponse.

This PR will fix this typo


ksinder:
Looks like this issue no longer exists if I'm reading the helpers file correctly, and I see `MentionRichTtextItemResponse` has a typo, but thank you for the PR!

PR #493, opened 2024-02-25
Update README.md with correct `cd` command

crosskayla:
**nitpick:** incorrect and / or misleading `cd` command

**explanation:** The slash `/` at the beginning of the path points to the root of the filesystem, so `cd /notion-sdk-js` tries to change to a directory that is directly under the root directory. That is not where the cloned repo would be based on the first `git clone` command

**reason for change:** could confuse some beginners 🤷‍♀️


ksinder (approved):
Amazing, thank you so much! LGTM

PR #500, opened 2024-04-09
update api endpoints with in_trash field in request and response

alexanderhzheng:
## Description

Update api endpoints with in_trash field in request and response
The goal is to eventually deprecate archived field.

Following guide: https://dev.notion.so/notion/Update-Notion-JS-SDK-57cce3061ed04a33b4663fee8f6f9ad1

- [ ] Unit test
- [ ] Client test
- [X] Integration test
- [ ] Browser (end-to-end) test
- [x] Manual test (explain below)
      verified locally

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -48,8 +48,6 @@ export type UserObjectResponse =
   | PersonUserObjectResponse
   | BotUserObjectResponse
 
-type StringRequest = string
-
 type SelectColor =
   | "default"
   | "gray"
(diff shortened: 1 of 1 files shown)


src/api-endpoints.ts
```diff
@@ -62,12 +60,6 @@ type SelectColor =
   | "pink"
   | "red"
 
-type SelectPropertyResponse = {
```
almawang, line 65: hm probably not related to your change, but it's not good that this got removed 🤔

alexanderhzheng: Yea I just ran `notion build-public-api-js-sdk --path=../notion-sdk-js/src/` which introduced a lot of other changes. I'm wondering if I should manually remove the changes I didn't introduce

alexanderhzheng: image

almawang (approved):

PR #506, opened 2024-05-03
Use isomorphic fetch

sv2dev:
Since `fetch` is stable in the current LTS version of Node, I don't see any good reason for using node-fetch or any other solution than the web standard.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 15.x, 16.x]
+        node-version: [18.x, 19.x, 20.x, 22.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


ksinder:
👋 Hi from Notion's API team! Thanks for bumping [this issue](https://github.com/makenotion/notion-sdk-js/issues/38#issuecomment-2792668696), looks like it's been open for a while. I don't have much context on this yet, though it looks like you have merge conflicts. I recently added a test that relies on us overriding the fetch passed to the `Client` constructor so we may need to restructure that if we go this route. Beyond that, I'll check with some folks internally to get caught up on our stance re: isomorphic fetch and see if I can get you an update here!

ksinder:
Getting rid of the `node-fetch` dependency seems great conceptually, but I think ideally we still want to allow passing a `fetch` constructor argument into the `Client` to support everyone's individual use-case, instead of only supporting isomorphic `fetch`. Is there any way to reference a common static type for fetch to get rid of the `node-fetch` dependency without regressing the customizability?

ksinder:
Amazing, thank you so much for the clean change and making all those changes, and for your patience! This looks good to me and worth shipping as part of our next version release.

ksinder (approved):

Issue #507, opened 2024-05-04
AnnotationResponse type not exported


jwanner83:
**Describe the bug**
The AnnotationResponse type, which is internaly used inside of any RichTextItemResponses isn't correctly exposed / usable as type. This would be easaly fixable with one word `export`.

The reason why this type should be exported is, because it helps when developing a ui element which displays the rich text response.

**To Reproduce**
Node version: 21.7.3
Notion JS library version: 2.2.15

I will create a pull request to fix this problem. Thats why I didn't add steps to reproduce.

PR #508, opened 2024-05-04
Export AnnotationResponse type (#507)

jwanner83:
As i promised in this issue: https://github.com/makenotion/notion-sdk-js/issues/507.

The pull request to add the `export` keyword to the `AnnotationResponse` type. Now, the type is available to import via `import { AnnotationResponse } from '@notionhq/client/build/src/api-endpoints'`

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -739,7 +739,7 @@ type VerificationPropertyResponse = {
     | null
 }
 
-type AnnotationResponse = {
+export type AnnotationResponse = {
   bold: boolean
   italic: boolean
   strikethrough: boolean


ksinder:
See the comment I just made in another PR: https://github.com/makenotion/notion-sdk-js/pull/523#issuecomment-2859275815
I believe once I publish `v3.0.1`, you should be able to extract this type from some of the rich text response object types that will be accessible from the package? Let me know if that doesn't work for you though and I can look into a solution! (The `api-endpoints.ts` file is generated so we prefer folks not change it directly since it'd get clobbered when my team runs the generation script for it!)

Issue #498, opened 2024-03-27
Inconsistent type: Cursor for parameters is string|undefined but for responses is string|null 


jcbages:
Report bugs here only for the Node JavaScript library.

If you're having problems using Notion's API, or have any other feedback about the API including feature requests for the JavaScript library, please email support at developers@makenotion.com.

**Describe the bug**
When calling something like `database.query` that receives a cursor as a parameter, its type is `string | undefined` as it's `start_cursor:?string`. However, the returned cursor from the response is of type `string | null`. This causes issues when looping to obtain multiple pages while reassigning the cursor variable as the initial typing of the cursor variable will conflict with one of these.

I checked the type definitions and this is the case for all calls that receive and return a cursor. This is easy to workaround but it degrades the typing experience a little.

**To Reproduce**
Node version:
Notion JS library version:
2.2.14

```
let databaseId = '<some_id>';
let entries = [];
let cursor: string | null = null;
let entries = [];

while (true) {
    const { results, has_more, next_cursor } = await this.notion.databases.query({
        database_id: databaseId,
        start_cursor: cursor
    });

    cursor = next_cursor;
    entries.push(...results);

    if (!has_more) {
        break;
    }
}

return entries;
```

**Expected behavior**
Expected: No typing issues
Got: Type 'string | null' is not assignable to type 'string | undefined'.

If we change the type of the cursor to be `string | undefined` then the error happens at `cursor = next_cursor` saying: `Type 'string | undefined' is not assignable to type 'string | null'.`

**Additional context**
Happy to make a PR for this :)

Sho-ki:
I made a PR for this issue. #509

PR #510, opened 2024-05-05
fix start_cursor request type (issue:#498)

Sho-ki:
Issue on https://github.com/makenotion/notion-sdk-js/issues/498

## Description
This PR resolves a type mismatch issue encountered when using the start_cursor in database.query.
The current implementation leads to a conflict between string | undefined and string | null in various scenarios involving the cursor.

## Changes Made
Updated the type for start_cursor in QueryParameters to accept both null and undefined.

## Expected Behavior:
With this change, you will no longer face typing issues when assigning next_cursor from a query response to a variable.

```typescript
  let databaseId = "<some_id>"
  let entries = []
  let cursor: string | null = null

  while (true) {
    const { results, has_more, next_cursor }: QueryDatabaseResponse =
      await notion.databases.query({
        database_id: databaseId,
        start_cursor: cursor,
      })

    cursor = next_cursor  // Now this works without typing issues
    entries.push(...results)

    if (!has_more) {
      break
    }
  }

  return entries
```

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -166,7 +166,7 @@ export default class Client {
     const url = new URL(`${this.#prefixUrl}${path}`)
     if (query) {
       for (const [key, value] of Object.entries(query)) {
-        if (value !== undefined) {
+        if (value !== undefined && value !== null) {
           if (Array.isArray(value)) {
             value.forEach(val =>
               url.searchParams.append(key, decodeURIComponent(val))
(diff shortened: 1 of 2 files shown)


corradopetrelli:
I also noticed this inconsistency. UP! 😄

ksinder:
Thank you! Sorry for the belated response & for the SDK issue; we've finally looped back to this and I'm planning on shipping https://github.com/makenotion/notion-sdk-js/pull/723 in the next minor version to address/supersede this PR. Appreciate the contributions :)

Issue #505, opened 2024-04-30
`request_id` missing on types `GetPageResponse`, `QueryDatabaseResponse` and `GetDatabaseResponse`


borisghidaglia:
**Describe the bug**
`request_id` missing on types `GetPageResponse`, `QueryDatabaseResponse` and `GetDatabaseResponse`

Indeed, it is included in the response but does not appear in types.
Similar to #487

**To Reproduce**
Node version: v21.7.3
Notion JS library version: 2.2.15

- retrieve a db using `client.pages.retrieve`, `client.databases.retrieve` or `client.databases.query` and save this data as a TS const somewhere.
- add the `satisfies` operator like and replace X with the appropriate type.
```ts
const notionData = {...} satisfies X; 
```
You will get this error (as an example with `GetPageResponse`): 
`Object literal may only specify known properties, and 'request_id' does not exist in type 'GetPageResponse'`

**Expected behavior**
`satisfies X` on the data should work as the functions mentioned above return a `Promise<X>`.

Issue #487, opened 2024-02-04
Missing `request_id` type in `OauthTokenResponse`


MohammedEsafi:
## Description:

Currently, the `OauthTokenResponse` does not include a field for the `request_id` type, even though this field is included in the actual response. this absence leads to a type error in a specific use case where I encounter the issue while utilizing that field.

![Screenshot 2024-02-04 at 21 16 46](https://github.com/makenotion/notion-sdk-js/assets/47392041/513b6551-4320-4f0f-b0b8-a1ec87b6fb85)

To address this issue, I have to cast the token method response. however, I believe this should be resolved within the library itself. I plan to open a PR to address this issue.

![Screenshot 2024-02-04 at 21 32 12](https://github.com/makenotion/notion-sdk-js/assets/47392041/3a8486ce-bfce-49a5-b0ef-ffaa3c2727ad)

ksinder:
This was shipped a while ago via https://github.com/makenotion/notion-sdk-js/pull/565/files#diff-e9db95173437635a114462cd206975f9c276db80d16fd842e7f8955f5e527b4d and is available in the SDK now for OAuth APIs, tysm!

PR #515, opened 2024-05-29
Issue #505 and issue #487.

bansalji02:
The specified issues discuss about how the request_id is not being returned in the specified functions. So, Adding the request_id in the return type of these functions would resolve the issue and allow the users to access the same.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -11582,6 +11582,7 @@ export type OauthTokenResponse = {
       }
     | { type: "workspace"; workspace: true }
   duplicated_template_id: string | null
+  request_id:string
 }
 
 export const oauthToken = {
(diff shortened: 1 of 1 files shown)


bansalji02:
@seratch PLease review and approve

bansalji02:
@aoberoi Please review and approve this
@blackmad

ksinder:
Thanks! Unfortunately we can't accept direct edits to the `src/api-endpoints.ts` file, since it's auto-generated from Notion's latest public API spec and doesn't sync back. We've recently shipped an improvement to surface `request_id` on the OAuth endpoints in #565, and are going to release v5.1.0 soon which also includes `request_id` in the logs of error responses. Can you clarify (in https://github.com/makenotion/notion-sdk-js/issues/505) the use case for programmatic access to `request_id` on the remaining endpoints?

Issue #519, opened 2024-07-03
Inconsistent Typescript Type Exports


adam26davidson:
This is not exactly a bug but more of a request for a quality of life enhancement for typescript

I am finding that with Typescript in strict mode, it would be nice to have access to more (or maybe just all?) of the types defined in `api-endpoints.d.ts`, and in general, have complex types split out into exported component types.

As an example, when updating a page, I am able to build a request as follows, which is nice because I can build the request object dynamically while ensuring the request type conforms to `UpdatePageParameters`:
``` typescript
const request: UpdatePageParameters = { page_id: task.id, properties: {} };
const properties = request.properties;
if (!properties) {
  throw new Error('Properties not found');
}

properties[taskSchema.status.key] = { status: { name: taskSchema.status.options.notCompletedInTime } };
properties[taskSchema.queue.key] = { select: { name: taskSchema.queue.options.archived } };

await this.notion.pages.update(request);
```
First of all, I can't do this sort of thing for calls to 'pages.create()' since the type used in that request, `CreatePageBodyParameters` is not exported from `api-endpoints.d.ts`, forcing me to use any for the type of the properties object and forgo typechecking

Second of all, it would be nice to have access to an exported "UpdatePageProperties" type here so that I don't have to check that `properties` exists in this code - currently the definition of that object is a part of `UpdatePageParameters` and can't be instantiated independently (maybe there is a typescript trick i am not aware of to do this but having a type would be more ergonomic in any case)

aschleck:
Agreed, it'd be amazing to have all the types exported.

In the meantime, a fun workaround. You can probably do this for input types as well.

```
// never call this function, it's used purely to extract the types
async function dummy(notion: NotionClient) {
  for await (
      const result of iteratePaginatedAPI(notion.databases.query, {database_id: ''})) {
    if (isFullPageOrDatabase(result)) {
      return result;
    }
  }
  throw new Error('oh no');
}

type PageObjectResponse = Awaited<ReturnType<typeof dummy>>;
```

PR #523, opened 2024-08-07
fix: export item response types from src/index.ts

JoshuaKGoldberg:
Sibling to #507 / #519: it looks like of odd that we're importing types from the long `@notionhq/client/build/src/api-endpoints.js`. This should let consumers import from the package itself:

```ts
import type { RichTextItemResponse } from "@notionhq/client";
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,9 @@
+export type {
+  EquationRichTextItemResponse,
+  MentionRichTextItemResponse,
+  RichTextItemResponse,
+  TextRichTextItemResponse,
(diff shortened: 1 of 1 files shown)


ksinder:
Amazing idea, thank you! 🎉 Inspired by this, I've included this change (along with a few additional exports) in https://github.com/makenotion/notion-sdk-js/pull/562 and plan to publish this as `v3.0.1` today :) closing this PR since mine subsumes these changes

JoshuaKGoldberg:
Aha awesome, thank you!

PR #526, opened 2024-09-26
Feat: api-endpoints.ts - export Properties Params on createDatabase

IvanCCO:
Export Properties Params to create dynamic title name for properties.  To create databases with internalization, the field name must be dynamic, and to get TS power, I exported the type to get the TSrecommendation


IvanCCO:
image

PR #532, opened 2024-11-13
Update .env.example

jtsmith1287:
Incorrectly named environment variable as referenced in the code example.


ksinder (approved):
Nice, thank you! LGTM

PR #568, opened 2025-05-21
Updates node/nvm version required

typeoneerror:
`nvm use` failed as package is asking for 18 and docs and .nvmrc wanted 12/14 respectively.


ksinder (approved):
Thank you for fixing!

PR #570, opened 2025-05-23
Add `max_file_upload_size_in_bytes` field to User response

ksinder:
Sync the latest OpenAPI schema for the Notion Public API to `src/api-endpoints.ts`.

Key changes include:
- Add a new field `workspace_limits: { max_file_upload_size_in_bytes: number }` to the `bot` info shape in the User API response.
  - Extract the `bot` info object to a reusable type `BotInfoResponse` to reduce the overall file size.
- Use `TextRequest` instead of `StringRequest` for several property response types (e.g. `SelectPropertyResponse `) that now have an increased character limit in the API.

### DIFF
--- a/src/api-endpoints.ts
+++ b/src/api-endpoints.ts
@@ -16,28 +16,29 @@ type EmptyObject = Record<string, never>
 
 export type PartialUserObjectResponse = { id: IdRequest; object: "user" }
 
+type BotInfoResponse = {
+  owner:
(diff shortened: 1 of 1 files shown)


seemaullal (approved):
makes sense to me and better than playing binary search with different file sizes to figure out the limit

mquan (approved):

PR #584, opened 2025-06-02
Clean up CreateComment and rich text content param types

ksinder:
- Sync the latest generated `src/api-endpoints.ts` from Notion's codebase to the JS SDK.
  - In this case, there are no semantic changes, but refactors and additional docstrings/comments in the Create Comment API body parameters based on recent internal refactors.


ksinder:
Including this as part of https://github.com/makenotion/notion-sdk-js/pull/590

PR #606, opened 2025-08-29
Add Notion ID extraction functions to helpers

riwaht:
## Description

This adds small helper functions to extract Notion IDs (database, page, block) from URLs.  
It makes it easier to work with Notion links by always giving back the correct UUID, even when URLs contain extra parameters like `?v=` for views.  

- `extractNotionId()` – core function  
- `extractDatabaseId()`, `extractPageId()`, `extractBlockId()` – simple wrappers  
- Works with common URL formats (`notion.so`, `notion.site`, shared links, block anchors, etc.)  
- Returns `null` if no valid ID is found  
- No breaking changes, just extra helpers  

---

 [X] Added unit tests in `test/id-extraction.test.ts`  
  - Covers database URLs with view params  
  - Page share links (`?p=`)  
  - Block fragments (`#block-...`)  
  - Compact and hyphenated UUIDs  
  - Invalid inputs return `null`  

[X] Manually tested with my own Notion databases, pages, and blocks  

---

## Screenshots

**Example:**

```ts
const dbId = extractDatabaseId("https://www.notion.so/workspace/MyDB-abc123...?v=viewid456");
// -> "abc123de-f456-7890-1234-5678901234ab"

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -199,3 +199,107 @@ export function isMentionRichTextItemResponse(
 ): richText is RichTextItemResponseCommon & MentionRichTextItemResponse {
   return richText.type === "mention"
 }
+
+/**
+ * Extracts a Notion ID from a Notion URL or returns the input if it's already a valid ID.
+ * 
+ * Prioritizes path IDs over query parameters to avoid extracting view IDs instead of database IDs.
+ * 
+ * @param urlOrId A Notion URL or ID string
+ * @returns The extracted UUID in standard format (with hyphens) or null if invalid
+ * 
+ * @example
+ * ```typescript
+ * // Database URL with view ID - extracts database ID, not view ID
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/helpers.ts [hunk 1]
```diff
@@ -199,3 +199,107 @@ export function isMentionRichTextItemResponse(
...
+  }
+
+  // Look for block fragment in URL (#block-32chars or just #32chars)
+  const blockMatch = urlWithBlock.match(/#(?:block-)?([0-9a-f]{32})/i)
+  if (blockMatch && blockMatch[1]) {
+    return formatUuid(blockMatch[1])
+  }
+
+  return null
+}
```
ksinder, line 305: sorry to nitpick but would you be able to add the trailing newline for this file?

ksinder (commented):
Looks good! Mind completing the CLA, and then we can approve and merge? https://github.com/makenotion/notion-sdk-js/pull/606#issuecomment-3236826730

ksinder (approved):

PR #636, opened 2025-10-24
optimized pick() to eliminate intermediate array allocation

khresth:
## Description

Optimized the `pick()` function to improve performance by eliminating intermediate array allocations. The original implementation used `keys.map()` to create an array of `[key, value]` tuples, then converted it to an object via `Object.fromEntries()`. This created temporary data structures on every invocation.

The new implementation directly constructs the result object in a single loop, reducing memory allocations and iteration overhead. Since `pick()` is called twice per API request (for query params and body params), this optimization should improve performance across SDK operations.

```
bash
node node_modules\jest\bin\jest.js ./test
```

Test Results:

PASS test/id-extraction.test.ts
PASS test/helpers.test.ts
PASS test/Client.test.ts

Test Suites: 3 passed, 3 total
Tests:       24 passed, 24 total

## Screenshots

image


ksinder (approved):
ty!!

PR #638, opened 2025-11-03
Tidy up gitignore, add CLAUDE.md

michaelfromyeg:
## Description

Couple of nice-to-haves.

- [ ] Automated test (unit, integration, etc.)

Claude Code works fine on my machine, same files still `gitignore`d.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,28 +15,33 @@ Official Notion SDK for JavaScript - a TypeScript client library providing type-
 ## Development Commands
 
 ### Building
+
 ```bash
 npm run build          # Runs: npm run clean && tsc
 npm run clean          # Remove build/ directory
 ```
 
 ### Testing
+
 ```bash
 npm test               # Run all tests in test/
 ```
 
 To run a single test file:
+
 ```bash
 npx jest test/helpers.test.ts
 ```
 
 ### Linting and Formatting
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,11 +1,158 @@
...
 
 .vscode
+.zed
 
 .DS_Store
+
+# Claude Code
+settings.local.json
+
+### Node.gitignore (https://github.com/github/gitignore/blob/main/Node.gitignore)
```
michaelfromyeg: Just piggybacks off the linked one directly, which covers most common build artifacts / is generally good to have for a Node project.

CLAUDE.md
```diff
@@ -0,0 +1,165 @@
+# CLAUDE.md
```
michaelfromyeg: If nothing else, this is great documentation so we should include it. It'd be nice to build an AGENTS.md, and Copilot file as well. Looking at open-source tools that do this.

ksinder: I thiiink we have Copilot instructions somewhere in this repo already but yeah your general thinking here sgtm, appreciate the amazing ownership here

michaelfromyeg: We do! And yeah, once I figure out something that can build out both, I'll consolidate. Figured this was a good first step.

PR #664, opened 2026-02-01
Add automatic retry support with exponential backoff

ksinder:
## Description

- Implements comprehensive automatic retry mechanism for the Notion SDK with exponential backoff and retry-after header support.
- Retries transient errors (429, 500, 503) with configurable behavior.
- Refactors the ~150-line `request()` method into focused private methods.
- Converts all interfaces to types and upgrades dev dependencies (Jest 28→29, eslint 7→8, @typescript-eslint 5→7).

**Test Details:**
- All 55 unit tests passing with Jest 29 fake timers
- Refactored retry tests use realistic timing scenarios instead of hacks
- New test validates maxRetryDelayMs cap
- Full build and lint suite passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,17 +140,19 @@ The `Client` supports the following options on initialization. These options are
 | `baseUrl`   | `"https://api.notion.com"` | `string`       | The root URL for sending API requests. This can be changed to test with a mock server.                                                                       |
 | `logger`    | Log to console             | `Logger`       | A custom logging function. This function is only called when the client emits a log that is equal or greater severity than `logLevel`.                       |
 | `agent`     | Default node agent         | `http.Agent`   | Used to control creation of TCP sockets. A common use is to proxy requests with [`https-proxy-agent`](https://github.com/TooTallNate/node-https-proxy-agent) |
-| `retry`     | `{ maxRetries: 3 }`        | `RetryOptions` | Configuration for automatic retries on rate limits (429) and server errors (500, 503). See [Automatic retries](#automatic-retries) below.                    |
+| `retry`     | `{ maxRetries: 2 }`        | `RetryOptions` | Configuration for automatic retries on rate limits (429) and server errors (500, 503). See [Automatic retries](#automatic-retries) below.                    |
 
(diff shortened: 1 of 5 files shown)


src/Client.ts
```diff
@@ -244,74 +306,246 @@ export default class Client {
...
+        if (!isNotionClientError(error)) {
+          throw error
+        }
+
+        this.logRequestError(error, attempt)
+
+        if (attempt < this.#maxRetries && this.canRetry(error)) {
+          const delayMs = this.calculateRetryDelay(error, attempt)
+          this.log(LogLevel.INFO, "retrying request", {
+            method,
```
mquan: Should we restrict retries to idempotent method (GET) only? If the problem is a transient network issue retrying can result in undesirable duplicate side effects?

README.md
```diff
@@ -132,14 +132,53 @@ You may also set a custom `logger` to emit logs to a destination other than `std
...
+  },
+})
+```
+
+To disable automatic retries:
+
+```js
+const notion = new Client({
+  auth: process.env.NOTION_TOKEN,
+  retry: { maxRetries: 0 },
```
mquan: is it also possible to disable retries with `retry: false`? May be we should just support one approach

src/Client.ts
```diff
@@ -244,74 +306,246 @@ export default class Client {
...
+      return { requestId: obj.request_id }
+    }
+    return {}
+  }
+
+  /**
+   * Determines if an error can be retried based on its error code.
+   */
+  private canRetry(error: unknown): boolean {
+    if (!APIResponseError.isAPIResponseError(error)) {
```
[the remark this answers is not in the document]
ksinder: yes, they should be! https://github.com/makenotion/notion-sdk-js/blob/4c56b4af43943733a1c6be47a5324a0ea09bca8c/src/errors.ts#L278-L302 -- the tests in the PR make me pretty confident this works. Can you clarify your question? are you saying they shouldn't be, or just confirming they are?

mquan (approved):
LGTM

Issue #666, opened 2026-02-14
Silent parameter dropping in pick() causes hard-to-debug issues


dimasriat:
## Problem

The SDK's internal `pick()` function in `helpers.js` silently drops any body parameters not listed in `api-endpoints.js`. There is no warning, no error — the parameter is simply ignored. Combined with the frequent need for `as any` casts (since TypeScript types don't always cover all valid API parameters), this creates a debugging nightmare where code appears correct but silently does nothing.

### Example 1: `databases.update` with `archived: true`

```typescript
// Developer expects this to archive a database — it silently does nothing
await client.databases.update({
  database_id: "...",
  archived: true,  // silently dropped — not in bodyParams
} as any);
```

`databases.update` bodyParams only include `in_trash`, not `archived`. But `pages.update` accepts BOTH `archived` and `in_trash`. A developer familiar with `pages.update` will naturally try `archived` on `databases.update` — and it silently fails.

### Example 2: `databases.create` with top-level `properties`

```typescript
// Developer expects this to create a database with custom columns — properties are silently dropped
await client.databases.create({
  parent: { type: "page_id", page_id: "..." },
  title: [{ type: "text", text: { content: "My DB" } }],
  properties: {  // silently dropped — not in bodyParams
    Status: { select: {} },
  },
} as any);
```

The correct field is `initial_data_source: { properties: { ... } }`, but there's no feedback when using the wrong structure.

## Impact

In our project, these silent drops caused **three separate multi-hour debugging sessions**:

1. `databases.create` with top-level `properties` → database created with only the default `Name` column, all subsequent `pages.create` calls failed with "property X does not exist"
2. `databases.update` with `archived: true` → cleanup function appeared to succeed but databases were never trashed, causing stale data in tests
3. Both issues produced errors far downstream from the actual bug, making root cause analysis extremely difficult

## Proposed Fix

Add a warning when `pick()` encounters unknown parameters. This is a minimal change in `src/helpers.ts`:

```typescript
function pick(base, keys) {
  const entries = keys.map((key) => [key, base?.[key]]);
  const unknownKeys = Object.keys(base ?? {}).filter(k => !keys.includes(k) && !pathOrQueryParams.includes(k));
  if (unknownKeys.length > 0) {
    console.warn(`@notionhq/client warn: unknown body parameters ignored: ${unknownKeys.join(", ")}`);
  }
  return Object.fromEntries(entries.filter(([, value]) => value !== undefined));
}
```

[...]

PR #667, opened 2026-02-14
Add warnings for unknown parameters passed to API methods

dimasriat:
## Summary

- Adds a `warnUnknownParams()` method to `Client` that detects when callers pass parameters not recognized by an endpoint's definition
- Warns via the existing `LogLevel.WARN` logger (respects user-configured log levels and custom loggers)
- Adds `getUnknownParams()` utility in `utils.ts` for testability
- Applied to all non-OAuth API endpoint methods (30 methods)

## Problem

The SDK's internal `pick()` function silently drops any body parameters not listed in `api-endpoints.ts`. Combined with the frequent need for `as any` casts (since TypeScript types don't always cover all valid API parameters), this creates a debugging nightmare.

**Example**: `databases.update` accepts `in_trash` but NOT `archived` (while `pages.update` accepts both). Calling `databases.update({ database_id, archived: true })` silently does nothing — no error, no warning.

See #666 for full details and reproduction steps.

## What the warning looks like

```
@notionhq/client warn: unknown parameters were ignored {
  unknownParams: [ 'archived' ],
  knownParams: [ 'database_id', 'parent', 'title', 'description', 'is_inline', 'icon', 'cover', 'in_trash', 'is_locked' ]
}
```

## Design decisions

- **Uses existing logger** — respects `logLevel` and custom `logger` options, so it can be silenced
- **`utils.ts` stays generic** — `getUnknownParams()` is a pure function, easy to test
- **OAuth endpoints excluded** — they extract `client_id`/`client_secret` manually outside the endpoint definition, which would cause false positives
- **Non-breaking** — warnings only appear at `WARN` level (default), existing behavior is unchanged

## Test plan

- [x] `npm run build` compiles without errors
- [x] Existing tests unaffected (pre-existing failures are Node.js version issues, not related to this change)
- [ ] Manual verification: passing `archived: true` to `databases.update` now logs a warning

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@notionhq/client",
-  "version": "5.9.0",
+  "version": "5.10.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@notionhq/client",
-      "version": "5.9.0",
+      "version": "5.10.0",
       "license": "MIT",
       "devDependencies": {
         "@types/jest": "29.5.14",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@notionhq/client",
-  "version": "5.9.0",
+  "version": "5.10.0",
   "description": "A simple and easy to use client for the Notion API",
   "engines": {
     "node": ">=18"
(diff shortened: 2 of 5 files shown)


ksinder (approved):
thank you so much!

PR #751, opened 2026-07-25
Surface edge-proxy blocks as a diagnosable error

ksinder:
## Description

When a request is rejected at the network edge, the response may be HTML rather than a Notion error body. The SDK already falls back to UnknownHTTPResponseError, but the generic message and raw HTML do not make the failure easy to diagnose.

This adds response diagnostics to the existing HTTP error types:

- ray_id is populated from the Cloudflare response header whenever available.
- request_id falls back to the Notion response header when an unrecognized response body does not expose one.
- An unrecognized response with a Ray ID and no Notion request ID is identified as edge-generated. HTTP 403 responses note that a network security rule may be responsible; 5xx responses do not make that causal claim.
- Well-formed Notion API errors still take precedence, and origin-generated responses such as OAuth errors keep the existing generic fallback message.

The shared response-header reader also makes Retry-After lookup case-insensitive for plain-object headers and preserves array-valued headers. No new error class is introduced, so the public NotionClientError union is unchanged.

Covered edge-generated 403 and 5xx responses, origin-generated OAuth and standard API errors, responses without proxy metadata, and standard, mixed-case, and array-valued header shapes. Also confirmed against an origin-generated OAuth error that it retains the generic message while exposing both response identifiers.

## Screenshots

N/A


src/errors.ts
```diff
@@ -332,6 +374,35 @@ export function buildRequestError(
...
+  const { status, contentType, ray_id } = args
+  const base = `Request to Notion API failed with status: ${status}`
+  if (ray_id === undefined) {
+    return base
+  }
+  const contentTypeNote =
+    contentType !== undefined ? ` (content-type: ${contentType})` : ""
+  return (
+    `${base}. The response was not generated by the Notion API${contentTypeNote}; ` +
+    `it was returned by Notion's edge proxy, which usually means the request was ` +
```
mquan: **[High] `cf-ray` doesn't identify edge-generated responses — every `api.notion.com` response carries one**

`api.notion.com` is behind Cloudflare, so `cf-ray` is on origin-generated responses too. Since `ray_id !== undefined` is the only gate, this message now fires for any error whose body isn't a parseable Notion error — not just HTML block pages.

The OAuth endpoints are a live instance: `{"error":"invalid_client","request_id":"..."}` has no `code`/`message`, so `parseAPIErrorResponseBody` returns `undefined` and we land here. Verified against prod on both `/v1/oauth/token` and `/v1/oauth/introspect`:

```
HTTP/2 401
content-type: application/json; charset=utf-8
cf-ray: a20c7d43582b6ad1-SJC
x-notion-request-id: 9c6f8f10-3107-435b-babe-87bec4d57877

{"error":"invalid_client","request_id":"7b656e0b-..."}
```

So `client.oauth.token()` with a wrong client secret now reports that the response "was not generated by the Notion API" and "was blocked by a network security rule" — every clause false, and it routes the integrator to Notion support for a config error. Same for any API error code outside `APIErrorCode`, and for Cloudflare 5xx (520/522/524) where the security-rule claim is also wrong.

The fix is in that header dump: `x-notion-request-id` can't be on a Cloudflare block page, since the request never reached origin. Gate on its **absence** instead of `cf-ray`'s presence (optionally also requiring a non-JSON content-type), and drop the causal claim for 5xx.

[written by Claude Code]

src/errors.ts
```diff
@@ -253,21 +281,35 @@ export function isHTTPResponseError(
...
+   */
+  readonly ray_id: string | undefined
+
   constructor(args: {
     status: number
     message: string | undefined
     headers: SupportedResponse["headers"]
     rawBodyText: string
   }) {
+    const ray_id = getResponseHeader(args.headers, "cf-ray")
```
mquan, line 303: **[Medium] `request_id` stays `undefined` even though the header carries it**

The constructor now reaches into `args.headers` for `cf-ray` and `content-type` but still hardcodes `request_id: undefined` below. `x-notion-request-id` is present on exactly the responses this PR targets as unhelpful, and it — not the Ray ID — is the identifier support uses for anything that actually reached the API. Populating it makes `UnknownHTTPResponseError` actionable for the origin-generated case, which is the majority case.

[written by Claude Code]

test/test-utils.ts
```diff
@@ -76,6 +76,28 @@ export function mockResponse(
...
+/**
+ * Creates a mock Response with a verbatim body and headers, for responses the
+ * Notion API did not generate (e.g. an edge proxy's HTML block page).
+ */
+export function mockRawResponse(args: {
+  status: number
+  body: string
+  headers?: Record<string, string>
+}): Response {
+  const headers = new Headers()
```
mquan, line 87: **[Medium] The one behavior fix this PR claims has no test coverage**

The description claims `getResponseHeader` fixes case-insensitive matching for plain-object headers, but nothing tests it. `mockResponse`, this new `mockRawResponse`, and the inline mocks in `Client.test.ts`/`helpers.test.ts` all build `new Headers()`, so execution never enters the plain-object branch — and there's no `test/errors.test.ts`. A direct unit test on `getResponseHeader` (`{"Retry-After": "120"}`, `{"CF-RAY": "..."}`) would cover it cheaply.

[written by Claude Code]

src/errors.ts
```diff
@@ -178,6 +178,34 @@ export function validateRequestPath(path: string): void {
...
+  }
+
+  if ("get" in headers) {
+    const headersObj = headers as { get: (name: string) => string | null }
+    return headersObj.get(name) ?? undefined
+  }
+
+  const lowerName = name.toLowerCase()
+  for (const [key, value] of Object.entries(headers)) {
+    if (key.toLowerCase() === lowerName && typeof value === "string") {
```
mquan: **[Low] Plain-object branch now drops array-valued headers, silently losing `retry-after`**

The `typeof value === "string"` filter rejects `string[]`. Node's `IncomingHttpHeaders` shape — what a non-`fetch`-native adapter is likely to return, and why `headers` is typed `unknown` in `fetch-types.ts` — can carry `{"retry-after": ["120"]}`. Before, the cast returned the array and `parseInt(["120"], 10)` coerced to `120`, so the server's backoff was honored; now it's `undefined` and `calculateRetryDelay` falls through to jittered exponential backoff, ignoring an explicit `Retry-After` on a 429. Taking the first element of an array would keep that working.

[written by Claude Code]

src/errors.ts
```diff
@@ -253,21 +281,35 @@ export function isHTTPResponseError(
 export class UnknownHTTPResponseError extends HTTPResponseError<ClientErrorCode.ResponseError> {
   readonly name = "UnknownHTTPResponseError"
 
+  /**
+   * Value of the `cf-ray` response header, when present. A response carrying a
+   * Ray ID but no Notion error body was produced by Notion's edge proxy rather
+   * than by the API itself — typically a security rule blocking the request
+   * before it reached Notion. Quote this ID when contacting Notion support.
+   */
+  readonly ray_id: string | undefined
```
mquan: **[Low] `ray_id` is only on `UnknownHTTPResponseError`, not `APIResponseError`**

The Ray ID is a support identifier for the connection, equally useful on ordinary API errors (e.g. debugging intermittent `service_unavailable`). Both classes extend `HTTPResponseError` and `getResponseHeader` is already available there, so lifting `ray_id` to the base class would avoid an asymmetric surface callers have to type-narrow around.

[written by Claude Code]

mquan (commented):
[written by Claude Code]

The goal is right and the shape is good (no new error class, Notion error body still wins). One blocking issue: `cf-ray` is present on *every* `api.notion.com` response, including origin-generated ones, so it can't gate the "not generated by the Notion API" claim. Details inline.

Residual blind spot: the HTML fixture is hand-built with only `content-type` and `cf-ray`, so no test can catch the misclassification. A pair of fixtures — one with `x-notion-request-id` (origin), one without (edge) — would pin the distinction.

mquan (approved):

PR #763, opened 2026-08-13
[SDK] Fix partial response type guards

notion-software-factory:
#### Problem

`isFullDataSource` and `isFullDatabase` checked only the resource `object` discriminator. Partial API responses use the same discriminator, so both helpers returned `true` and incorrectly narrowed partial objects to full response types.

#### Solution

Require the full-only `title` field in both structural guards, matching the approach already used by `isFullPage` and `isFullBlock`. Add regression coverage for partial and full data source/database shapes, including compile-time use of full-only fields after narrowing.

#### Result

Partial data source and database responses now return `false`, while full responses return `true`. The TypeScript build confirms the exported predicates still narrow to `DataSourceObjectResponse` and `DatabaseObjectResponse`.

#### Details

The existing wiki data-source query fixture now includes `title: []` because it represents a full child data-source row.

Closes BUG-324367.

[Code change explainer](https://app.dev.notion.com/p/3bbb35e6e67f8126b953f008195f01e3)

[Agent conversation](https://app.dev.notion.com/agent/32cb35e6e67f804c84820092ec038c49?wfv=activity&at=3bbb35e6e67f819e845900a9d52b8fe3)

Validated with `npm run build`, the focused helper test suite, all 106 Jest tests, and `npm run lint` (Prettier, ESLint, and cspell).

## Screenshots

Not applicable; this change has no UI impact.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -378,7 +378,7 @@ export function isFullPage(
 export function isFullDataSource(
   response: ObjectResponse
 ): response is DataSourceObjectResponse {
-  return response.object === "data_source"
+  return response.object === "data_source" && "title" in response
 }
 
 /**
@@ -387,7 +387,7 @@ export function isFullDataSource(
 export function isFullDatabase(
   response: ObjectResponse
 ): response is DatabaseObjectResponse {
-  return response.object === "database"
+  return response.object === "database" && "title" in response
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ksinder:
sorry, didn't realize the software factory/agent produced this PR! I only saw it file the ticket and assumed the separate repo caused it to run into permissions issues or something

closing this in favor of mine: #764

PR #779, opened 2026-08-25
fix: allow column_list/column as children of single-level-nestable blocks

### LINKED ISSUE #575: "callout" blocks should have children
**Describe the bug**
Right now callout blocks are defined under the `BlockObjectRequestWithoutChildren` type, while in reality they do have children, and they should be listed in the `BlockObjectRequest` union type.
This can be verified by retrieving a callout block with nested children from the API.

This inaccuracy is also present in the [docs](https://developers.notion.com/reference/block#callout), that don't show the `children` property for callout blocks

**To Reproduce**
Node version:
Notion JS library version: 3.1.2

ShayanMuhammad-CS:
## Summary
Fixes #575.

The original report claimed `callout` blocks were missing from `BlockObjectRequest` entirely — that part was already fixed on `main` and a prior attempt (#613) was rightly closed since it couldn't reproduce a compile error for that case.

However, a real bug was left unresolved further down the same issue thread: [@Fental's comment](https://github.com/makenotion/notion-sdk-js/issues/575#issuecomment-2543129331) posted a real API response showing a `callout` block containing a nested `column_list` → `column` → `paragraph`. That shape does not type-check on `main` today:

```
error TS2322: Type '"column_list"' is not assignable to type '"code" | "file" | "toggle" | ... | "synced_block"'.
```

## Root cause
`BlockObjectWithSingleLevelOfChildrenRequest` (`src/api-endpoints/common.ts`) models the children of blocks that support one level of nesting — `callout`, `toggle`, `quote`, `bulleted_list_item`, `numbered_list_item`, `heading_1-4`, `paragraph`, `template`, `synced_block`. That union is missing `column_list`/`column` entirely, even though the top-level `BlockObjectRequest` union supports them fine. So a column layout could be created at the root of a page, but not nested inside any of those container blocks — despite the Notion API allowing it.

## Fix
Added `column_list`/`column` variants to `BlockObjectWithSingleLevelOfChildrenRequest`, mirroring the existing top-level `ColumnListRequest`/`ColumnWithChildrenRequest` definitions but with children capped one level deeper (`BlockObjectRequestWithoutChildren`), consistent with how every other nestable block type in that union reduces its child depth (e.g. `quote`, `toggle`).

## Test plan
- [x] `npm run build` (`tsc`) — no errors
- [x] `npm run lint` — no new warnings/errors introduced by this change
- [x] `npm test` — 130/130 passing
- [x] Added `test/block-nesting-types.test.ts`, a type-level regression test (via `ts-jest`) that:
  - asserts a callout with a nested `column_list`/`column`/`paragraph` layout now compiles
  - asserts a block still cannot go one level deeper than allowed (e.g. a `quote` nested inside a `column` still correctly fails to compile)


ShayanMuhammad-CS:
recheck

ShayanMuhammad-CS:
recheck

ShayanMuhammad-CS:
recheck

ShayanMuhammad-CS:
Hi @ksinder — following up since you were involved in the discussion on #575 and the earlier attempt (#613). This PR fixes the underlying bug from that thread: \`column_list\`/\`column\` were missing from \`BlockObjectWithSingleLevelOfChildrenRequest\`, so a callout with a nested column layout (as shown in the API response @Fental posted in the issue) didn't type-check even though the API supports it.

Build/lint/tests all pass locally, and I added a regression test (\`test/block-nesting-types.test.ts\`). Would appreciate a review/workflow approval when you have a chance. Thanks!

PR #783, opened 2026-08-27
[SDK] Check compatibility before refactors ship

ksinder:
## Description

Check the exports and types available from `@notionhq/client` against the PR base. The check catches removed exports and fields, narrower class method inputs, and constructor access changes. It ignores private implementation changes and checks protected members used by subclasses.

Unchanged and patch versions must pass. A deliberate minor or major version increase allows planned contract changes. This is a conservative check of the package entry point, not a runtime test or a check of all deep imports. It uses the existing TypeScript compiler and adds no dependencies.

This pass also enforces braces across handwritten TS/JS, removes an unchecked byte-index cast, and includes JavaScript in linting. Shared agent rules now live in `AGENTS.md`. The README is a reference; `CONTRIBUTING.md` is a how-to guide. The docs distinguish the public SDK from Notion's internal generator. The README uses the verified public sticker-logo asset.

- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm test -- --runInBand --silent`: 166 tests passed
- [x] Full SDK compatibility check against a separate checkout of current main
- [x] Regression tests for all of Quan's findings, plus overloads, nested results, protected members, and generic async iterators
- [x] Checks against three open sync PRs, each compared with its own merge base

The sync-PR sample produced these results:

| PR | Result |
| --- | --- |
| #784: add optional `logical_key` | Passed. |
| #773: remove `stream.end.last_sequence` | Rejected the response type and affected Client exports. |
| #772: remove tool-call fields | No change to the package entry point. A direct check of `src/api-endpoints.ts` rejected both affected response types. |

The sample PRs were not changed. The logo URL returned HTTP 200 with `image/svg+xml` and passed XML validation.

## Screenshots

Not applicable.


test/checkCompatibility.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (!declaration) throw new Error(`Missing declaration: ${symbol.name}`)
+    return checker.getTypeOfSymbolAtLocation(symbol, declaration)
+  }
+
+  function compatible(
+    name: string,
+    currentType: ts.Type,
+    previousType: ts.Type
+  ) {
+    if (!checker.isTypeAssignableTo(currentType, previousType)) {
```
mquan: **[Medium] Compatibility check misses removed optional response fields**

Removing an optional field leaves the new type assignable to the previous type, so the check passes even though existing property access stops compiling. For example, removing `SearchResponse.request_status` would pass both the export and Client.search return-type comparisons, while breaking consumers that read `response.request_status`

test/checkCompatibility.ts
```diff
@@ -0,0 +1,253 @@
...
+  const signatures = (constructors.length ? constructors : [undefined]).map(
+    constructor =>
+      ts.factory.createFunctionTypeNode(
+        declaration.typeParameters,
+        constructor?.parameters ?? [],
+        ts.factory.createKeywordTypeNode(ts.SyntaxKind.VoidKeyword)
+      )
+  )
+  return [
+    ...members,
```
mquan, line 369: **[Medium] Compatibility check misses narrowed public class methods**

TypeScript compares method parameters bivariantly, and this transformation leaves methods unchanged while normalizing only constructors. Narrowing the argument accepted by the public Client.request method would therefore pass, even though existing calls using the removed input no longer compile.

test/checkCompatibility.ts
```diff
@@ -0,0 +1,253 @@
...
+function isPrivateMember(member: ts.ClassElement): boolean {
+  return (
+    Boolean(member.name && ts.isPrivateIdentifier(member.name)) ||
+    Boolean(
+      ts.canHaveModifiers(member) &&
+        ts
+          .getModifiers(member)
+          ?.some(
+            modifier =>
+              modifier.kind === ts.SyntaxKind.PrivateKeyword ||
```
mquan: **[Medium] Making the Client constructor private can pass compatibility**

A private or protected constructor is removed before constructor analysis, leaving an implicit no-argument constructor in the rewritten declaration. Changing the current public Client constructor to private can therefore compare as compatible while breaking every `new Client()` caller; the same filtering also hides protected-member removals from subclasses.

mquan (approved):

test/checkCompatibility.ts
ksinder: Fixed in 5875c46. The check now verifies property presence in addition to assignability, including optional fields in nested results, promises, arrays, and union members. Added regression tests for removed optional response fields and tests that keep additive fields valid.

test/checkCompatibility.ts
ksinder: Fixed in 5875c46. Public class methods are compared as function properties so strict input checks apply. Intersections preserve overloads, and generic parameters stay intact. Tests cover narrower public, static, and protected methods, removed overloads, and valid wider inputs.

test/checkCompatibility.ts
ksinder: Fixed in 5875c46. Private and protected constructors keep their visibility in the declarations used for comparison. A separate subclass view checks protected members without treating the two checkouts as different class identities. Tests cover restricted constructors, removed or narrowed protected members, unchanged protected members, and protected-to-public changes.

PR #791, opened 2026-09-02
Warn when the client holds a token inside a browser page

ksinder:
## Description

Adds a `dangerouslyAllowBrowser` client option and a "Browser usage" README section. When the client is built with `auth` inside a browser page, it logs one warning: anyone who can load the page can read the token and act as the integration. Passing `dangerouslyAllowBrowser: true` confirms the choice and silences the warning. OpenAI and Anthropic use the same option name in their SDKs.

This is the SDK half of the change that makes `api.notion.com` send CORS headers (https://github.com/makenotion/notion-next/pull/336692). Merge after that server change is live, because the README now states the API sends CORS headers.

Four new Jest cases cover: warns once in a browser page with a token, quiet with the option set, quiet without a token, quiet outside a browser. `npm run build`, `npm run lint`, and the Client test suite pass locally.

## Screenshots

None. No UI change.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -114,6 +114,13 @@ export type ClientOptions = {
    * Set to false to disable retries entirely.
    */
   retry?: RetryOptions | false
+  /**
+   * Confirms that you mean to hold a Notion token inside a browser page, and
+   * silences the warning the client logs in that case. Anyone who can load the
+   * page can read the token and act as your integration, so only set this for
+   * pages that nobody else can open.
+   */
+  dangerouslyAllowBrowser?: boolean
 }
 
 type FileParam = {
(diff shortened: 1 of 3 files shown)


ksinder (commented):
[written by Claude Code] Fresh-eyes review by a subagent found no blockers. Three follow-ups were applied in the second commit: the browser check also covers web and service workers, per-request `auth` now warns once per client, and the README notes that jsdom test runners trigger the warning and how to quiet it. Build, lint, and 86 Client tests pass. GitHub blocks approving your own PR, so this is a comment rather than an approval.

src/Client.ts
```diff
@@ -114,6 +114,13 @@ export type ClientOptions = {
    * Set to false to disable retries entirely.
    */
   retry?: RetryOptions | false
+  /**
+   * Confirms that you mean to hold a Notion token inside a browser page, and
+   * silences the warning the client logs in that case. Anyone who can load the
+   * page can read the token and act as your connection, so only set this for
+   * pages that nobody else can open.
+   */
+  dangerouslyAllowBrowser?: boolean
```
JasonPuglisi, line 123: How do other providers handle this flag - do they also warn if omitted, or block completely? Maybe we should block?

PR #795, opened 2026-09-03
Fix/654 generic formula property types

### LINKED ISSUE #654: Use generic type for FormulaPropertyItemObjectResponse
This issue is a suggestion.

**Describe the bug**
When dealing with formula property, I noticed : 
```
type FormulaPropertyResponse = StringFormulaPropertyResponse | DateFormulaPropertyResponse | NumberFormulaPropertyResponse | BooleanFormulaPropertyResponse;
```
This way, if I have 2 properties in my datasource (one string, other a date), typescript cannot give me the corresponding type.

**To Reproduce**
Node version: 24.11.1
Notion JS library version: 5.6.0

- Have at least one formula property in database ;
- Make a datasource query with notion sdk ;
- Retrieve formula value

**Expected behavior**
Currently I use this workaround :
```
{
          id: result.id,
          slug:
            "string" in properties.slug.formula &&
            properties.slug.formula.string
              ? properties.slug.formula.string
              : ""
}
```
I have to manually check if formula contains the wanted value. 

We can solve this by doing this way :
```
export type FormulaPropertyItemObjectResponse<T extends string | Date | number | boolean> = {
    type: "formula";
    formula: FormulaPropertyResponse<T>;
    object: "property_item";
    id: string;
};
type FormulaPropertyResponse<T extends string | Date | number | boolean> = T extends string ? StringFormulaPropertyResponse : T extends Date ? DateFormulaPropertyResponse : T extends number ? NumberFormulaPropertyResponse : T extends boolean  ? BooleanFormulaPropertyResponse : never;
````

Usage :
```
export type NotionArticleDTO = {
  slug: FormulaPropertyItemObjectResponse<string>;
  title: TitlePropertyItemObjectResponse;
  publishedAt: FormulaPropertyItemObjectResponse<Date>;
  status: StatusPropertyItemObjectResponse;
  category: SelectPropertyItemObjectResponse;
};
```

What do you think about that ?

### PR DESCRIPTION
**Pull request by ayyYoSam**
# Description

Fixes: #654

Add generic type support to `FormulaPropertyItemObjectResponse` so consumers can narrow formula property responses to the expected value type.

The existing default behavior is preserved, including support for unsupported formula types, so this change does not introduce a breaking change.

* [x] Automated test (unit, integration, etc.)
* [ ] Manual test (provide reproducible testing steps below)

TypeScript compilation was verified with:

```bash
npx tsc --noEmit
```

A dedicated type test was added to verify that string, number, date, and boolean formula responses can be narrowed using the generic type parameter, and that mismatched formula types are rejected by TypeScript.

## Screenshots

N/A

### DIFF
--- a/src/api-endpoints/pages.ts
+++ b/src/api-endpoints/pages.ts
@@ -87,19 +87,23 @@ export type FilesPropertyItemObjectResponse = {
   id: string
 }
 
-export type FormulaPropertyItemObjectResponse = {
+type FormulaValue = string | DateResponse | number | boolean
+
+type FormulaPropertyResponse<T extends FormulaValue = FormulaValue> =
+  | (T extends string ? StringFormulaPropertyResponse : never)
+  | (T extends DateResponse ? DateFormulaPropertyResponse : never)
+  | (T extends number ? NumberFormulaPropertyResponse : never)
+  | (T extends boolean ? BooleanFormulaPropertyResponse : never)
+  | UnsupportedFormulaPropertyResponse
+
+export type FormulaPropertyItemObjectResponse<
+  T extends FormulaValue = FormulaValue,
+> = {
   type: "formula"
-  formula: FormulaPropertyResponse
+  formula: FormulaPropertyResponse<T>
   object: "property_item"
   id: string
 }
-
-type FormulaPropertyResponse =
-  | StringFormulaPropertyResponse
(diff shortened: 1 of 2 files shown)


ayyYoSam:
Update :)

Manual tests were performed checking for generic type narrowing, backward compatible usage, and rejection of formula types that are incompatible with the expected TypeScript type.

PR #804, opened 2026-09-15
[SDK] Add typed Skills and plugin directory methods

notion-software-lab:
#### Attribution
- Notion user: Geoffrey Litt (glitt@makenotion.com)
- Notion thread: [View thread](https://app.dev.notion.com/chat?t=3dcb35e6e67f80b7966000a9b7f6e94d)

Companion to [makenotion/notion-next#347452](https://github.com/makenotion/notion-next/pull/347452). Add typed `client.plugins.list()`, `client.plugins.retrieve()`, and `client.skills.retrieve()` methods for the Agent Skills API, plus their public parameter and response exports. Retrieve methods return the signed archive URL; they do not download or extract files.

The endpoint types and wrappers are generated from the upstream PR with the `agent-skills` resource selection. The plugin facade encodes opaque tag IDs as one URL segment. Unrelated generated changes to shared types are excluded.

### Why the data-source helper changes are included

These helpers previously accepted the entire `Client` type. Adding `plugins` and `skills` therefore also made those unrelated helper parameters require the new namespaces, causing the existing compatibility check to fail for `iterateDataSourceTemplates`, `collectDataSourceTemplates`, `iterateAllDataSourceRows`, and `collectAllDataSourceRows`.

The helpers only use `client.dataSources`, so their parameter is now `Pick<Client, "dataSources">`. This widens the accepted input without changing runtime behavior: full clients still work, and callers do not need the new Skills methods to use data-source helpers. The helper tests exercise this smaller input. The compatibility check passes against SDK main with the package version unchanged; reverting these changes would restore the failure.

### Usage

```ts
import { Client } from "@notionhq/client"

const notion = new Client({
  auth: process.env.NOTION_API_KEY,
  notionVersion: "2026-03-11",
})

const plugins = await notion.plugins.list({ page_size: 100 })
const plugin = await notion.plugins.retrieve({
  id: "b2d0a1f4-9c3e-4c07-9a1d-5f6e7c8b9a01",
})
const skill = await notion.skills.retrieve({
  id: "b55c9c91-384d-452b-81db-d1ef79372b75",
})
```

### Release coordination

Land the upstream mappings/generator fix in #347452 before this generated SDK change. Release the SDK through a separate reviewed version bump; this PR does not change the package version or publish to npm. Keep the developer docs' working `request()` examples until that release, then switch them to the named methods and identify the minimum SDK version.

The installed local package was checked with intercepted HTTP responses, not live Notion credentials. Live API authorization and archive contents remain unverified in this SDK PR.

## Screenshots

Not applicable.


test/Client.test.ts
```diff
@@ -100,6 +100,121 @@ describe("Notion SDK Client", () => {
...
+          method: "GET",
+          headers: expect.objectContaining({
+            authorization: "Bearer ntn_override",
+            "Notion-Version": "2026-03-11",
+          }),
+        })
+      )
+    })
+
+    it("retrieves a skill archive URL without downloading the archive", async () => {
```
geoffreylitt: hm what does "without downloading the archive" mean, i think we can remove that from the test name here

src/Client.ts
```diff
@@ -650,6 +651,18 @@ export default class Client {
    * Notion API endpoints
    */
 
+  public readonly plugins = {
+    list: this.#endpointMethods.plugins.list,
+    // Plugin IDs can be tag IDs containing URL delimiters, not just UUIDs.
```
geoffreylitt: move this comment right above the id: line?

geoffreylitt (approved):

mquan (approved):

